PR #2, opened 2022-04-01
added simple graphics stuff

nickwn:
peepee


TheCrossBoy (changes requested):
Should chariot-client/Cargo.lock and chariot-core/Cargo.lock be having these changes committed? I think the .gitignore was edited after these commits happened, so I want to confirm before the commit goes through

TheCrossBoy:
other than that tho, looks neat

i'm about 78% sure i understand what's going on with it, which is (probably) a good sign

possibly just a note for the future, but I think we're probably going to want to in the end we aren't going to want to expose the "true" event loop to the end game client, so that we can properly deal with lots of different stuff wanting to "hook" into the game's core loop, but that can be dealt with later and is mostly an aside

also - we may want to add timing to the render passes (and possibly a way to hook into/remove the hooked fns), that way instead of explicitly running the code loop every time, our renderer just handles that behind the scenes, and we just have to call "hey i want this to render in the future" or "hey deal with input in this way", and it'll do that without explicitly calling said functions on every game loop manually

not sure if we want to take that route in the end but it may make our lives easier down the line, especially if we want multiple passes of stuff or multiple sources dealing with input, but we can talk about that later

TheCrossBoy (dismissed):

KNOXDEV:
I'm gonna wait for Nick's highly anticipated "game loop feature" before I review. Until then I'm converting this back into a draft.

nickwn:
A lot of this new stuff is untested but it should be good enough to merge and play with input handling. I added an ecs ([specs](https://docs.rs/specs/latest/specs/)) for scene management but it might do more harm than good since it makes things kinda convoluted for a small game. We can tack it on to the input handling discussion this sunday. Also let me add Cargo.toml rn so the builds stop failing.

TheCrossBoy (approved):
turns out the highly anticipated feature was well worth the wait

chariot-client/src/application.rs
```diff
@@ -0,0 +1,222 @@
...
+impl<'a> StaticMeshRenderSystem<'a> {
+	fn new(renderer : &'a mut Renderer, resources : &'a ResourceManager) -> Self {
+		Self {
+			renderer : renderer,
+			resources : resources
+		}
+	}
+}
+
+impl<'a, 'b> specs::System<'b> for StaticMeshRenderSystem<'a> {
```
TheCrossBoy, line 174: here's where that odd indentation was, I think this ones in spaces and everything else is tabs

chariot-client/src/drawable.rs
```diff
@@ -0,0 +1,288 @@
...
+		let material = resources.materials.get(&self.material)
+			.expect("invalid material handle");
+
+		let vertex_buffers_with_ranges = 
+			static_mesh.vertex_buffers.iter().zip(static_mesh.submeshes[self.submesh_idx].vertex_ranges.iter());
+
+		let mut bind_group_refs = vec![&self.xform_bind_group];
+		bind_group_refs.extend(material.bind_groups.values());
+		RenderItem::Graphics { 
+            pass_name: material.pass_name.as_str(),
```
TheCrossBoy, line 90: some stuff here is spaces as well (edit: throughout the function)

chariot-client/src/main.rs
```diff
@@ -1,11 +1,89 @@
+use std::mem;
+use wgpu::util::DeviceExt;
+use winit::{
```
TheCrossBoy, line 3: s p a c e s

TheCrossBoy: its most of this file except for like 3 lines in main which is odd but ye

chariot-client/src/renderer.rs
```diff
@@ -0,0 +1,890 @@
+use core::panic;
```
TheCrossBoy, line 1: this file is entirely with spaces

chariot-client/src/shader.wgsl
```diff
@@ -0,0 +1,16 @@
+struct Locals {
```
TheCrossBoy, line 1: also spaces here

TheCrossBoy (changes requested):
left markers where i found spaces, but all the content looks good

KNOXDEV:
Just run rustfmt on the project and be done with it. No need to review for it.

KNOXDEV:
@nickwn I agree with your trepidation towards including a full-blown ECS. I'm not sure what everybody's familiarity is with this architecture, but unless we're used to and attached to it I think it will cause complexity problems.

It would be more useful if we were doing physics simulation on the client as well, but since the code bases are separate, the interop between networking and specs and physics and specs is going to become ugly fast. Also, we don't have the complexity required to benefit from parallel processing. 

For those reasons, I would say drop Specs. Thoughts?

KNOXDEV:
I would also recommend some organizational changes. Things like ResourceManager being pulled out of application.rs (there seems to be a few different functionalities contained in there) as well as breaking apart the structs and functions in renderer.rs. 

The line count of renderer.rs is staggering. Great work getting so much done, but if you could pull some of the functionality into external modules it would do amazing things for readability. We can also chalk this up to my lack of graphics exposure, but I'm struggling to understand some of the functionality with the current number of comments. If you have the time to add some context-provided comments, especially in renderer.rs, it would go a long ways. If you don't find the time its not the end of the world.

Thanks again Nick. Looking forward to the next canary build.

TheCrossBoy (approved):

TheCrossBoy:
removing my changes since that got fixed w the github actions

PR #3, opened 2022-04-01
networking bootstrap

KNOXDEV:
don't look im not ready yet

### DIFF
--- a/chariot-client/src/game.rs
+++ b/chariot-client/src/game.rs
@@ -1,0 +1,25 @@
+use std::net::TcpStream;
+use chariot_core::packets::ServerUpdatingPacket;
+
+pub struct GameClient {
+    connection: TcpStream
+}
+
+impl GameClient {
+    pub fn new(ip_addr: String) -> GameClient {
+        let connection = TcpStream::connect(&ip_addr)
+            .expect("could not connect to game server");
+        // disable the Nagle algorithm to allow for real-time transfers
+        connection.set_nodelay(true)
+            .expect("could not turn off TCP delay");
+        println!("game client now listening on {}", ip_addr);
+        GameClient {
+            connection
+        }
+    }
+
+    pub fn ping(&self) {
+        ServerUpdatingPacket::Ping.write_packet(&self.connection)
+            .expect("failed to send ping packet")
+    }
+}
--- a/chariot-server/src/game.rs
+++ b/chariot-server/src/game.rs
@@ -1,11 +1,17 @@
+use std::{io, thread};
+use std::collections::VecDeque;
+use std::io::Read;
 use std::net::{TcpListener, TcpStream};
-use std::thread;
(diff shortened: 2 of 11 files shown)


KNOXDEV:
big big big picture order of events here:
1. implement a simple ping/pong server-client interaction using the standard libraries' [`net` crate](https://doc.rust-lang.org/std/net/index.html), here is [an example](https://gist.github.com/andelf/8668088)
2. Design the custom TCP protocol in depth.
a. Will involve sending game state deltas from Server to Client using bincode serialization.
b. Will involve sending input events from Client to Server.
c. more to be seen after designing

KNOXDEV:
### What is finished?
1. Server -> Client connection. First, run the server. Then, run the client. You'll see the client send a ping packet to the server and the server will receive it.
2. The server logic loop has been completely scaffolded. Physics and other simulation elements can be added easily.
3. More packets containing arbitrary state can be added fairly easily.

### What still needs to be done?
* Quality of Life, ie, when a client disconnects, the game should notice and pause until they reconnect.
* Server -> Client transmission is done, but Client -> Server transmission needs to be taken care of. Since the work on graphics is going to heavily affect the structure of the client, I added the minimum code necessary to test the connectivity, but a similar packet-queuing / processing structure that currently exists in server will need to be added to client. Thankfully, pretty much anybody will be able to do this work using the existing structure as a reference.
* Adding more packets that sync useful data, but hopefully this is a self-evident task as well.

chariot-client/src/game.rs
```diff
@@ -0,0 +1,25 @@
...
+
+pub struct GameClient {
+    connection: TcpStream
+}
+
+impl GameClient {
+    pub fn new(ip_addr: String) -> GameClient {
+        let connection = TcpStream::connect(&ip_addr)
+            .expect("could not connect to game server");
+        // disable the Nagle algorithm to allow for real-time transfers
```
trulyronak, line 12: what is the Nagle algorithm?

chariot-server/src/game.rs
```diff
@@ -0,0 +1,117 @@
...
+
+            // wait until server tick time has elapsed
+            let remaining_tick_duration = max_server_tick_duration.checked_sub(start_time.elapsed())
+                .expect("server tick took longer than configured length");
+            thread::sleep(remaining_tick_duration);
+        }
+    }
+
+    // blocks the primary loop if we don't have the minimum players
+    fn block_until_minimum_connections(&mut self) {
```
trulyronak, line 49: would it be possible to potentially also make the server send out the current number of players connected here?

trulyronak (approved):
super solid this all p much makes sense, had a few questions + one ask to see if that's possible

chariot-server/src/game.rs
KNOXDEV: eventually

PR #4, opened 2022-04-04
Add minimal framework for particle simulation

alistairgray42:
Stage 1 of a physics engine is a particle simulation. Particles have position, velocity, and mass, but not dimensions. This PR completes this by adding functionality for the following things:

- Calculating the value of all forces we care about (see functions from `sum_of_forces_on_object` for the list) 
- Allowing for rotation (as discussed, rotation is free and we don't really deal with angular velocity in a way that corresponds to real-world physics)
- Moving the object each time step based on the sum of the forces and other data from the previous step

The assumptions that this simulation makes that we'll invalidate later include:
- that objects don't have dimensions ofc
- that chairs are always on the ground
- that the ground is flat (physics gets a lot harder when you start tilting lol)

Includes tests for straight-line forces as well as for turning - all tests pass.

Stage 2 will be giving objects bulk / implementing collision detection.
Other potential future stages:
- Parametrization of physics (to account for different chair properties)
- The third dimension.
- more?

Formulas used in calculation of forces are taken from [the link Jack suggested](https://asawicki.info/Mirror/Car%20Physics%20for%20Games/Car%20Physics%20for%20Games.html)


chariot-server/src/physics/constants.rs
```diff
@@ -0,0 +1,12 @@
+pub const GRAVITY_COEFFICIENT: f64 = 1.0;
```
alistairgray42, line 1: So all the constants in here are meant to be changed once we actually start playing, but they should at least be in the right kind of ballpark for now so that physics generally works as expected (e.g. friction moving an object back isn't stronger than forward motion, so things go in the right direction) - i think these values are fine but yeah that's something in particular to convince yourself about

alistairgray42:
Requested review from Aaron for Rust conventions, anyone else feel free to review as well

chariot-server/src/physics/constants.rs
alistairgray42: same comment applies but these have moved - see `chariot-core/settings`

trulyronak:
Note - you're committing changes to Cargo.lock, likely from prior stages of the repo when we allowed Cargo.lock files - you'll need to remove those changes from the repo so it doesn't change anything in the repo

chariot-core/src/physics_object.rs
```diff
@@ -0,0 +1,35 @@
...
+pub enum RotationStatus {
+	InSpinClockwise,
+	InSpinCounterclockwise,
+	NotInSpin
+}
+
+pub struct PhysicsProperties {
+	pub position: DVec3,
+	pub velocity: DVec3,
+	pub linear_momentum: DVec3, // redundant with velocity; both are used for convenience's sake
```
KNOXDEV, line 27: Redundancy can introduce inconsistency. Have one value and add a method to quickly calculate the other.

chariot-core/src/settings.rs
```diff
@@ -4,19 +4,40 @@ use serde::Deserialize;
...
 pub struct Settings {
+	// Networking
     pub port: String,
     pub server_address: String,
     pub server_tick_ms: u64,
-    pub player_amount: usize
+    pub player_amount: usize,
+
+	// Physics
+	pub gravity_coefficient: f64,
```
KNOXDEV, line 14: can you add some comments describing what these are each used for?

alistairgray42:
> Quick comment: does it make sense to have the physics simulation structs in chariot-core if only the server is going to simulating physics?
> 
> I would put everything in chariot-server until we know we have a concrete need for something to also be in chariot-core.

The client needs some of this information (e.g. position, angle facing) to be able to render properly - it doesn't need things like velocity or momentum, but splitting that off into a server-only struct while still sending this stuff back and forth seemed inelegant to me. All of this is very much state, and in the minimal MVP all state gets sent back and forth a lot.

alistairgray42:
> Quick comment: does it make sense to have the physics simulation structs in chariot-core if only the server is going to simulating physics?
> 
> I would put everything in chariot-server until we know we have a concrete need for something to also be in chariot-core.

The client needs some of this information (e.g. position, angle facing) to be able to render properly - it doesn't need things like velocity or momentum, but splitting that off into a server-only struct while still sending this stuff back and forth seemed inelegant to me. All of this is very much state, and in the minimal MVP all state gets sent back and forth a lot.

KNOXDEV:
> The client needs some of this information (e.g. position, angle facing) to be able to render properly - it doesn't need things like velocity or momentum, but splitting that off into a server-only struct while still sending this stuff back and forth seemed inelegant to me. All of this is very much state, and in the minimal MVP all state gets sent back and forth a lot.

I would argue physics and networking can and should be separate concerns. If you don't like the idea of translating physics structures (which include information about velocity) into networking structures (which concern position and orientation) then perhaps we should take the position and orientation information, put them into their own structure (say, `Entity`) in chariot-core, but then have `PhysicsProperties` *contain* an Entity alongside all of the other physics stuff. When it comes time to transmit the important networking stuff, we only send the nested Entity struct.

How does that sound?

chariot-server/src/physics/mod.rs
```diff
@@ -0,0 +1,211 @@
...
+use glam::DVec3;
+
+use chariot_core::GLOBAL_CONFIG;
+use chariot_core::physics_object::PhysicsProperties;
+use chariot_core::physics_object::EngineStatus;
+use chariot_core::physics_object::RotationStatus;
+
+/* Given a set of physical properties, compute and return what next tick's
+ * physics properties will be for that object */
+pub fn do_physics_step(previous_props: &PhysicsProperties, time_step: f64) -> PhysicsProperties{
```
KNOXDEV, line 12: maybe convert this to an `impl fn` on PhysicsProperties?

KNOXDEV: same for the other functions in this file

chariot-server/src/physics/mod.rs
```diff
@@ -0,0 +1,211 @@
...
+		angular_velocity: angular_velocity,
+		mass: previous_props.mass,
+		engine_status: previous_props.engine_status,
+		rotation_status: previous_props.rotation_status,
+		unit_steer_direction: previous_props.unit_steer_direction,
+	};
+}
+
+
+pub fn sum_of_forces_on_object(object: &PhysicsProperties) -> DVec3
```
KNOXDEV, line 37: also, if you have specific sources for these formulas, including the links via comments would be delightful

KNOXDEV (changes requested):
Here's my feedback. Can you take a look? 

Thanks a bunch Alistair, this looks really solid. The tests are a nice touch.

alistairgray42:
> I would argue physics and networking can and should be separate concerns. If you don't like the idea of translating physics structures (which include information about velocity) into networking structures (which concern position and orientation) then perhaps we should take the position and orientation information, put them into their own structure (say, Entity) in chariot-core, but then have PhysicsProperties contain an Entity alongside all of the other physics stuff. When it comes time to transmit the important networking stuff, we only send the nested Entity struct.

I'm game to separate these concerns, but I think there are three categories of data here rather than two:
* Information sent from client to server - this is whether the player is accelerating / braking / turning (currently: `engine_status`, `rotation_status`)
* Information sent from server to client - position and orientation of vehicles (currently: `position`, `velocity`, `unit_steer_direction`; in the future, perhaps more to support the third dimension)
* Server-internal physics simulation information (currently: `mass`, `angular_velocity`; of course the simulation also uses data in the other categories)

I propose having, respectively, `PlayerControls`, `EntityLocation`, and `PhysicsProperties` structs: the client sends `PlayerControls`  to the server and receives `EntityLocation`s, while `PhysicsProperties` on the server contains both of those. `PlayerControls` and `EntityLocation` would live in core, while `PhysicsProperties` would only be on the server. (not married to those names by the way, naming things is hard)

Thoughts?

KNOXDEV:
> I propose having, respectively, `PlayerControls`, `EntityLocation`, and `PhysicsProperties` structs: the client sends `PlayerControls` to the server and receives `EntityLocation`s, while `PhysicsProperties` on the server contains both of those. `PlayerControls` and `EntityLocation` would live in core, while `PhysicsProperties` would only be on the server. (not married to those names by the way, naming things is hard)
> 
> Thoughts?

Good idea, lets do it.

* I think EntityLocation is good
* I would probably do PlayerInputs instead of PlayerControls
* I would maybe do PlayerEntity instead of PhysicsProperties

KNOXDEV (changes requested):
Can you rename chariot-core/src/physics_object.rs since the name isn't really applicable anymore? I'm thinking player_inputs.rs, and then sticking EntityLocation into another file as well if possible.

chariot-server/src/physics/mod.rs
```diff
@@ -0,0 +1,217 @@
...
+}
+
+fn rolling_resistance_force_on_object(object: &PlayerEntity) -> DVec3
+{
+	return object.velocity * object.mass * -1.0 * GLOBAL_CONFIG.rolling_resistance_coefficient;
+}
+
+
+#[test]
+fn test_accelerating() {
```
KNOXDEV, line 95: Tests should go in a tests module. They can either be in this file or in a tests.rs file in the same folder. 
https://doc.rust-lang.org/rust-by-example/testing/unit_testing.html

KNOXDEV:
@alistairgray42 passing back to you, when you're ready for the final check, just click this button to pass the review back to me:
image

KNOXDEV (commented):
@alistairgray42 finished looking at this, I love it. I noticed in physics/mod.rs, you have some tests in a test module but some are not. Was this intentional?

As soon as we get that cleared up, that'll be an approval for me. Great work.

KNOXDEV (approved):
i release you my sweet child

PR #13, opened 2022-04-09
implement client to server networking

KNOXDEV:


KNOXDEV:
We now have two way networking.

image
image

This pull request also contains a lot of reencapsulation on the networking side, but we have a really good level of code sharing between client and server now. Ready for review!

chariot-core/src/networking/connection.rs
```diff
@@ -0,0 +1,96 @@
+use std::collections::VecDeque;
+use std::io::{ErrorKind, Read, Write};
+use std::net::TcpStream;
+
+use super::Packet;
+
+pub struct Connection<T: Packet, V: Packet> {
```
trulyronak, line 7: seeking to learn: why `T` and `V` for the generic types of the incoming/outgoing packets?

chariot-server/src/game.rs
```diff
@@ -56,72 +59,30 @@ impl GameServer {
...
-        }
-    }
-
     // handle every packet in received order
     fn process_incoming_packets(&mut self) {
-        while let Some(packet) = self.incoming_packets.pop_front() {
-            match packet {
-                ServerUpdatingPacket::Ping => println!("Received a Ping packet!"),
+        for (i, connection) in self.connections.iter_mut().enumerate() {
+            while let Some(packet) = connection.pop_incoming() {
```
trulyronak, line 72: seeking to learn: what does `Some` do here?

trulyronak (approved):
this all looks good + super clean + readable, just had some questions for myself, but this looks good to go!

alistairgray42 (commented):
Mostly looks good to me! Question about testing: it seems at the moment you're doing ad-hoc integration testing which works, and I get that unit-testing networking code could definitely be a pain, but do we think that changes to this code are going to be frequent enough that automated testing would be worth it? (If so, [this crate](https://crates.io/crates/mockstream) I found might be useful) I could totally see an answer of "this demonstrably works as is, there won't be more complex functionality, and what we have won't change a lot, we don't need tests" being valid here fwiw

trulyronak:
oo fair point about tests - generally i've noticed with networking testing we just need to test the methods itself / it might be a pain to add testing for the networking.

@alistairgray42 were you gonna look into adding the networking tests?

chariot-core/src/networking/connection.rs
KNOXDEV: T and V are conventional generic parameter names, but I can call them anything, in particular probably something more descriptive can be arranged. 

It's just a legacy convention, see here for historical info https://stackoverflow.com/questions/44589035/where-does-the-t-u-v-convention-for-generic-type-params-come-from

chariot-server/src/game.rs
KNOXDEV: `while let` is a language construct shorthand that is actually a lot like a `match` clause inside of a `loop`.

In this usage, `pop_incoming` returns an `Option<Packet>`, which can be either `Some(Packet)` or `None()`.
We're basically saying keep calling `pop_incoming` until something other than `Some(Packet)` comes out, at which point, `break`. By using `while let`, we don't have to specify the `break` case, it's implied.

See here for more information: https://doc.rust-lang.org/rust-by-example/flow_control/while_let.html

KNOXDEV:
> Mostly looks good to me! Question about testing: it seems at the moment you're doing ad-hoc integration testing which works, and I get that unit-testing networking code could definitely be a pain, but do we think that changes to this code are going to be frequent enough that automated testing would be worth it? (If so, [this crate](https://crates.io/crates/mockstream) I found might be useful) I could totally see an answer of "this demonstrably works as is, there won't be more complex functionality, and what we have won't change a lot, we don't need tests" being valid here fwiw

Great question. Yes, I've written what essentially amounts to an integration test between client and server at the moment. At some point it will be replaced with the core logic of the game so it would indeed be nice to create a unit test of the components within `chariot-core` specifically.

Using TCP streams for this is fine, I think using mockstream might add a bit too much complexity. Good regression tests should evaluate functionality in a way that reflects the intended usage as closely as possible, and if we can just use TCP and what we've written already uses TCP, lets just use TCP.

I think anything more complicated than that for testing would bring diminishing value. I also think I shouldn't be the one to write these tests. Let's give the opportunity to somebody else. @trulyronak mentioned maybe being interested.

alistairgray42:
I can take a shot at writing tests here.

KNOXDEV:
> I can take a shot at writing tests here.

Go for it. I would suggest creating those in a new branch. I'm going to merge this since its been approved, but feel free to spin that up.

PR #14, opened 2022-04-10
Add client-server networking tests on localhost

alistairgray42:
as per title and comments on #13


KNOXDEV (approved):
Looks great to me. Thanks Alistair!

PR #15, opened 2022-04-10
Collisions

alistairgray42:
Add support for collisions to the physics engine.

This is Step 2 of the core physics functionality:

1. Particle simulation - forces acting on bulkless noninteracting particles on flat ground (#4)
2. This PR - gives objects bulk and allows them to collide.
3. Tie physics and networking together
4. Allow for parametrization of physics - not strictly necessary, but probably required for most of the fun audience interaction
5. Support for ramps and general verticality - not strictly necessary, depending on level design, but will make the game substantially more fun

Parts of this PR:

- [x] Give objects dimensions
- [x] Add methods for determining whether two objects are colliding
- [x] Thoroughly test those methods (suggestions for more tests / commits to add more are welcome!)
- [x] Apply forces to objects that are colliding based on the collision
- [x] Add physics steps to `simulate_game` in `game.rs` so that physics gets tied into networking


chariot-server/src/physics/collisions.rs
```diff
@@ -0,0 +1,230 @@
...
+// Check whether any portion of p1 is within p2
+fn check_collision(p1: &PlayerEntity, p2: &PlayerEntity) -> bool {
+    let p1_corners = p1.get_corners_of_hitbox();
+    let p2_corners = p2.get_corners_of_hitbox();
+
+    // Two steps: check if the y-value is fine (easy cause we're always
+    // spinning flat), and if it is then check whether (x, z) is within the 2d
+    // rectangle
+
+    for p1_corner in p1_corners {
```
alistairgray42, line 80: There's a pathological case which is likely not a practical problem but which passes this collision test: two cubes centered at the same location, one of which is rotated, say, 45 degrees relative to the other. This can be remedied by checking more points (the center of the object, center of the hitbox's edges) but I don't think that's necessary at the moment

chariot-server/src/physics/collisions.rs
```diff
@@ -0,0 +1,282 @@
...
+
+    return false;
+}
+
+fn are_players_colliding(p1: &PlayerEntity, p2: &PlayerEntity) -> bool {
+    return check_collision(&p1, &p2) || check_collision(&p2, &p1);
+}
+
+impl PlayerEntity {
+    pub fn collide_players(&self, other: &PlayerEntity, time_step: f64) -> Option<PlayerEntity> {
```
alistairgray42, line 123: Is `Option` a semantically good thing to use here? I often see it for, like, error handling, but this other usage made sense to me

chariot-server/src/game.rs
```diff
@@ -5,9 +5,17 @@ use std::time::{Duration, Instant};
 use chariot_core::networking::{ClientConnection, ClientUpdatingPacket, ServerUpdatingPacket};
 use chariot_core::GLOBAL_CONFIG;
 
+use crate::physics::player_entity::PlayerEntity;
+
 pub struct GameServer {
     listener: TcpListener,
     connections: Vec<ClientConnection>,
+
+    game_state: ServerGameState,
```
alistairgray42, line 14: This is very much a single idea (that may be far from best) for how to structure server-side state; suggestions welcome

KNOXDEV: It'll work for now, no foreseeable issues.

chariot-server/src/physics/collisions.rs
KNOXDEV: I think we can solve this issue and simplify your existing algorithm by using the industry standard Axis-Aligned Bounding Box technique instead. Basically, instead of rotating the bounding boxes, don't. It actually doesn't really matter that much in practice. 

https://developer.mozilla.org/en-US/docs/Games/Techniques/3D_collision_detection

chariot-server/src/physics/collisions.rs
KNOXDEV: `Option` should be used when a function will occasionally, in *normal* execution, need to indicate there is no return value.

Conversely, `Result` is used when the algorithm could have an *exceptional* error that the caller will need to handle.

I think that it is appropriate in this case to use `Option`.

To connect these concepts to another language like Java, `Option` is Rust's `null` and `Result` is Rust's checked exceptions.

chariot-server/src/physics/collisions.rs
```diff
@@ -0,0 +1,282 @@
...
+        assert!(!are_players_colliding(&high_cube, &low_cube)); // just for good measure
+    }
+
+    #[test]
+    fn test_collision_on_rotated_edges() {
+        // uwu w-wat if i was a cube with edge wength 10 (⁄ ⁄•⁄ω⁄•⁄ ⁄) centewed
+        // at the o-owigin and wotated 45 degwees (>ω^)
+        // a-and u (☆ω☆) wewe a cube with edge wength 10 a-awso wotated 45
+        // degwees (o･ω･o) but c-centewed 10sqwt(2) units away in the
+        // x-diwection Owo a-and we 😳 👉 👈 t-touched edges 🥺
```
KNOXDEV, line 254: this comment was corrupted somehow, please remove for common decency reasons

chariot-server/src/game.rs
```diff
@@ -81,7 +92,14 @@ impl GameServer {
     }
 
     // update game state
-    fn simulate_game(&mut self) {}
+    fn simulate_game(&mut self) {
+        for (index, player) in self.game_state.players.iter().enumerate() {
```
KNOXDEV, line 96: I see that you're doing a lot of cloning and copying when I think you could do all of this more efficiently with references.

For example, `self.game_state.players.iter_mut()` will allow you to iterate over mutable references to players and most likely do the same things.

From there, you could `filter` out the current player and recollect them into another vector for passing into `do_physics_step` without doing any additional cloning.

It's a little bit complicated but I would advise doing it. If you need any pointers just lmk

chariot-server/src/physics/mod.rs
```diff
@@ -6,16 +6,21 @@ use chariot_core::player_inputs::PlayerInputs;
...
 
 use player_entity::PlayerEntity;
 
 impl PlayerEntity {
     /* Given a set of physical properties, compute and return what next tick's
     	* physics properties will be for that object */
-    pub fn do_physics_step(&self, time_step: f64) -> PlayerEntity {
-        let forces = self.sum_of_forces_on_object();
-        let acceleration = forces / self.mass;
+    pub fn do_physics_step(
```
KNOXDEV, line 17: It's not top priority, but if you wanted to generalize this function to get passed in an arbitrary vector of bounding boxes that aren't necessarily players, that would be useful for future PRs where you'll need to handle player's colliding with terrain.

KNOXDEV (changes requested):
@alistairgray42 sending back to you. Great work on what you have, I'm giving some suggestions on the comments you left as well as some things I noticed. Let me know what you think.

chariot-server/src/physics/collisions.rs
```diff
@@ -0,0 +1,282 @@
...
+
+        // Angles are measured from the (1, 0, 0) axis
+        let heading = DVec2::new(
+            self.entity_location.unit_steer_direction[0],
+            self.entity_location.unit_steer_direction[2],
+        );
+        let zero_angle_vec = DVec2::new(1.0, 0.0);
+
+        // Since the range of arccos is [0, pi], we add the extra pi if turned around more than that
+        let theta;
```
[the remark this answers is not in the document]
alistairgray42: so true

chariot-server/src/physics/collisions.rs
```diff
@@ -0,0 +1,282 @@
...
+        if heading[1] >= 0.0 {
+            theta = heading.dot(zero_angle_vec).acos();
+        } else {
+            theta = heading.dot(zero_angle_vec).acos() + std::f64::consts::PI;
+        }
+
+        let center = DVec3::new(
```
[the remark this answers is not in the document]
alistairgray42: so true

chariot-server/src/physics/collisions.rs
alistairgray42: image
i'll have you know i paid good money for this comment

chariot-server/src/physics/mod.rs
alistairgray42: Yeah, I'd prefer to confine this PR to player-player collisions only at the moment

chariot-server/src/physics/collisions.rs
```diff
@@ -0,0 +1,257 @@
...
+        // reflectionally symmetrical over the X and Z axes, we can thus
+        // constrain (for the purposes of bounding-box calculation) the rotation
+        // angle to be in [0, pi/2].
+        let mut theta: f64 = heading.dot(zero_angle_vec).acos();
+        if theta > std::f64::consts::FRAC_PI_2 {
+            theta -= std::f64::consts::FRAC_PI_2;
+        }
+
+        // These two vectors and -1 times these two vectors define the four
+        // corners of the bounding box (source: trust me bro)
```
alistairgray42, line 40: i can draw a diagram for this if anyone needs convincing

chariot-server/src/physics/player_entity.rs
```diff
@@ -2,11 +2,16 @@ use chariot_core::entity_location::EntityLocation;
 use chariot_core::player_inputs::PlayerInputs;
 use glam::DVec3;
 
+#[derive(Copy, Clone, PartialEq)]
 pub struct PlayerEntity {
     pub velocity: DVec3,
     pub angular_velocity: f64, // in radians per time unit
     pub mass: f64,
 
+    pub x_size: f64,
```
KNOXDEV, line 11: is there a reason you didn't make this a DVec3 like the others? Idk if it would be better but I feel like it would be more consistent at least.

chariot-server/src/physics/player_entity.rs
```diff
@@ -2,11 +2,16 @@ use chariot_core::entity_location::EntityLocation;
 use chariot_core::player_inputs::PlayerInputs;
 use glam::DVec3;
 
+#[derive(Copy, Clone, PartialEq)]
```
KNOXDEV, line 5: I'm wondering why you've added Copy, Clone, and PartialEq to every struct in the physics engine. 

I tend to leave these underived since generally, Copying, Cloning and field-by-field Equality are operations that can be avoided in most cases without too much additional effort and some usage of references and borrowing. Moreover, if they can be avoided, they should be.

Additionally, by leaving them underived, the compiler will throw an error when you forget to pass by reference, which is a net good thing.

chariot-server/src/game.rs
```diff
@@ -81,7 +92,18 @@ impl GameServer {
...
 
     // update game state
-    fn simulate_game(&mut self) {}
+    fn simulate_game(&mut self) {
+        for player in self.game_state.players.iter() {
+            let others = self
+                .game_state
+                .players
+                .iter()
+                .filter(|&other_player| other_player != player)
```
KNOXDEV, line 101: I'm not sure if this equality check is a good way to go about this. Deriving PartialEq on the PlayerEntity struct means that the definition of equality is that every field in the struct is compared for equality to see if the overall structs are equal. I think you can see why this might be undesirable.

You can just add enumerate before the filter to filter by index, similar to how you did before. I think that would be a good way to go about this.

chariot-server/src/physics/collisions.rs
```diff
@@ -0,0 +1,257 @@
...
+    let z = v[1];
+
+    let sin_theta = theta.sin();
+    let cos_theta = theta.cos();
+
+    return DVec2::new(x * cos_theta - z * sin_theta, x * sin_theta + z * cos_theta);
+}
+
+impl PlayerEntity {
+    fn get_bounding_box_corners(&self) -> [[f64; 2]; 3] {
```
KNOXDEV, line 18: Is the name of this function still representative? It looks like this returns dimensions and not corners.

chariot-server/src/physics/collisions.rs
```diff
@@ -0,0 +1,257 @@
...
+        let min_x = self.entity_location.position.x + xs.into_iter().reduce(f64::min).unwrap();
+        let max_x = self.entity_location.position.x + xs.into_iter().reduce(f64::max).unwrap();
+
+        let min_z = self.entity_location.position.z + zs.into_iter().reduce(f64::min).unwrap();
+        let max_z = self.entity_location.position.z + zs.into_iter().reduce(f64::max).unwrap();
+
+        let min_y = self.entity_location.position.y - y_2;
+        let max_y = self.entity_location.position.y + y_2;
+
+        return [[min_x, max_x], [min_y, max_y], [min_z, max_z]];
```
KNOXDEV, line 67: This function returns min and max of all three dimensions. Is there any case where min_x would not simply be the same distance from the center as max_x but in the other direction?

I only ask because you compute min_x and max_x separately.

chariot-server/src/physics/collisions.rs
```diff
@@ -0,0 +1,257 @@
...
+    }
+}
+fn check_bounding_box_collisions(p1: &PlayerEntity, p2: &PlayerEntity) -> bool {
+    let bounding_box_1 = p1.get_bounding_box_corners();
```
KNOXDEV, line 71: So, I notice with this algorithm that over the course of a single physics loop, you compute the bounding box of each player `n` times (where `n` is the number of players). That makes each physics loop `n^2` in terms of the most computationally expensive part of the calculation.

It also has the side effect of making physics non-deterministic. That is to say, the order in which physics interactions are computed is dependent on their natural ordering of players in the PlayerEntity vector.

If instead you compute the AABB of every player once at the start of the physics loop and use those for every computation, you bring the algorithm back down to linear while also making physics independent of the uncontrollable player order.

chariot-server/src/physics/collisions.rs
```diff
@@ -0,0 +1,257 @@
...
+        } {
+            collision_dimensions[dimension] = true;
+        }
+    }
+
+    return collision_dimensions.iter().all(|&x| x);
+}
+
+fn are_players_colliding(p1: &PlayerEntity, p2: &PlayerEntity) -> bool {
+    return check_bounding_box_collisions(&p1, &p2) || check_bounding_box_collisions(&p2, &p1);
```
KNOXDEV, line 94: Does this really need to be computed twice here? I feel like this operation should be commutative.

chariot-server/src/physics/mod.rs
```diff
@@ -41,10 +46,21 @@ impl PlayerEntity {
...
             angular_velocity: angular_velocity,
             mass: self.mass,
+            x_size: self.x_size,
+            y_size: self.y_size,
+            z_size: self.z_size,
         };
+
+        for collider in potential_colliders {
+            if let Some(result) = post_physics_PlayerEntity.collide_players(&collider) {
+                post_physics_PlayerEntity = result;
```
KNOXDEV, line 56: I think Rust doesn't really play nicely with the pattern you're using here of "replacing" the entity position with a new one. I think you should have collision calculation functions be mutable functions on PlayerEntity so they can just modify the internal velocity data.

This is a bit of a more subjective piece of advice but in my experience it can be a lot cleaner. Do you have a strong reason for doing it this way?

KNOXDEV: This goes for most of the physics engine by the way, I notice you take a functional approach of having pure functions with inputs and outputs while avoiding state.

This is a good approach in a lot of languages but I think mutable functions are a lot more clear in Rust specifically.

KNOXDEV (changes requested):
Hey Alistair, loving the changes you made. Thanks a bunch, the AABB are for sure a much better way to go.

I'm still confused by several things because I'm not used to everybody's coding styles yet. I was hoping you could help me understand a few remaining places where the structure didn't click for me.

Thanks in advance Alistair!

chariot-server/src/physics/collisions.rs
alistairgray42: True! The previous version wasn't commutative, good catch.

chariot-server/src/physics/collisions.rs
alistairgray42: Yes, this is true

chariot-server/src/physics/player_entity.rs
alistairgray42: Sure

chariot-server/src/physics/mod.rs
alistairgray42: I understand the point here, and I definitely tend too much towards functionalbrain. I've replaced this specific instance you're committing on in the most recent commits, but am elsewhere still doing roughly functional things. In particular, I think this is justified in `simulate_game`: when colliding with multiple other objects simultaneously, each collision is going to cause a change to the velocity, but we want to use the original properties of all objects to calculate the new velocities, so we can't change velocity in-place without copying at some point.

chariot-core/src/entity_location.rs
```diff
@@ -2,7 +2,7 @@ use glam::DVec3;
 
 // EntityLocation gets sent back from the server to the client to give it
 // information on the results of the simulation == where to render players
-#[derive(Copy, Clone, PartialEq)]
+#[derive(Copy, Clone)]
```
alistairgray42, line 5: Would like to not have to derive Copy and Clone for these, but I think it's necessary(?) if `do_physics_step` doesn't mutate state, which I think it shouldn't do.

KNOXDEV: If `do_physics_step` just returns a new location, it shouldn't need to copy anything, unless I'm misunderstanding.

chariot-server/src/physics/mod.rs
```diff
@@ -131,9 +154,12 @@ mod tests {
             velocity: DVec3::new(2.0, 0.0, 1.0),
             angular_velocity: 0.0,
             mass: 10.0,
+
+            size: DVec3::new(10.0, 10.0, 10.0),
         };
 
-        props = props.do_physics_step(1.0);
+        let dims = props.get_bounding_box_dimensions();
```
KNOXDEV, line 161: Based on how frequently this call gets made, I think you should store the bounding box inside of the PlayerEntity itself. That way you can reuse it in various functions without needing to either recompute it every time or pass it in as arguments constantly.

At the beginning of the loop, you just call `player.update_bounding_box` to bring the bounding boxes up to date with the position and rotation, then you can use them in the physics computation as `self.bounding_box`, etc.

For the record, this is a pretty common approach.

How does that sound?

KNOXDEV (changes requested):
Looking very good at this point. Just a couple last comments and you can take a look and let me know what you think. 

Overall, I think eventually we'll be looking at a refactor that decouples the physics engine from the concept of Players and has FreeBodyEntity be its own structure with most of the logic, but I think this will not be too difficult and can be safely deferred.

chariot-server/src/physics/mod.rs
alistairgray42: Yeah, I went back and forth on that (didn't want to have state get out of sync but tbh that's probably not much of a real concern). Probably is cleaner like this, yes

KNOXDEV (approved):
:)

PR #16, opened 2022-04-13
basic outline of ECS

caIcutec:
I did some research on ECS stuff and made a draft of the code. May need some help determining what needs to be a component, and how to connect them etc.

Basic idea: A world scene contains entities, which each may contain children and/or components. Each component has an update function which can be used for rendering / other things. Entity and World have update functions where they just call update on all their children / components

World
entities:
- Entity1
    - children:
        - Entity3 
    - components: 
        - PlayerController
        - Rigidbody
        - Drawable
- Entity2
    - components:
        - Drawable
- etc...

Things to do:
- [ ] Get a list of needed components
- [ ] Fill out update()s for components
- [ ] Write tests

### DIFF
--- a/chariot-client/src/scenegraph/lib.rs
+++ b/chariot-client/src/scenegraph/lib.rs
@@ -1,0 +1,47 @@
+struct World {
+
+}
+
+struct Entity {
+	components: Vec<dyn Component>,
+}
+
+trait Component {
+	fn get_name(&self) -> &str;
+	fn update(&self);
+}
+
+struct Rigidbody {
+	position: {f32, f32},
+	velocity: {f32, f32},
+	acceleration: {f32, f32}
+}
+
+impl Component for Rigidbody {
+	fn update(&self){
+		self.velocity += self.acceleration;
+		self.position += self.velocity;
+	}
+}
+
+impl Entity {
+	fn update(&self){
+		for component in self.components.iter_mut(){
+			component.update();
+		}
+	}
+	fn add_component(&self, component: dyn Component){
+		self.components.push(component);
+	}
+	fn get_component(&self, name: &str) -> dyn Component{
+		for component in self.components {
+			if component.get_name() == &str {
+				return component;
+			}
+		}
+	}
+}
+
+fn main(){
+
+}


chariot-client/src/scenegraph/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+impl Entity {
+	fn update(&self){
+		for component in self.components.iter_mut(){
+			component.update();
+		}
+	}
+	fn add_component(&self, component: dyn Component){
+		self.components.push(component);
+	}
+	fn get_component(&self, name: &str) -> dyn Component{
```
nickwn, line 36:
```
fn get_component<T>(&self) -> Option<&T>
    where
        T: Component + 'static,
{
    let id = TypeId::of::<T>();
    self.components
        .get(&id)
        .map(|c| c.as_any().downcast_ref::<T>().unwrap())
}
```

chariot-client/src/scenegraph/lib.rs
```diff
@@ -0,0 +1,47 @@
+struct World {
+
+}
+
+struct Entity {
+	components: Vec<dyn Component>,
+}
+
+trait Component {
```
nickwn, line 9: To make casting easier, add: 
```
fn as_any(&self) -> &dyn Any;
```
Using Any from std::any::Any. Then in the actual component impls, add:
```
fn as_any(&self) -> &dyn Any {
    self
}
```

nickwn (commented):
Ok good start and sorry for putting you on this, I think it was a little more complicated than I thought it would be. After these modifications everything should be easier. Lmk here if there are any details I forgot to cover. Hopefully adding the children stuff now shouldn't be too hard.

It's kind of a pedantic thing but usually when people talk about ECSs they are referring to a specific data-oriented structure which we aren't doing because it's a little bit much. I guess what we're doing is just called the standard object-oriented structure although it's stupid since they both have entities and components. Basically if you’re looking up ECS you might see something different

As far as testing, if you have more time it might be nice to build a simple scene graph in Application and change the render() method to do a DFS traversal and accumulate the parent transforms (Use drawable.update_xforms() to update the model matrix).

chariot-client/src/scenegraph/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+	fn get_name(&self) -> &str;
+	fn update(&self);
+}
+
+struct Rigidbody {
+	position: {f32, f32},
+	velocity: {f32, f32},
+	acceleration: {f32, f32}
```
alistairgray42, line 17: I believe that, since velocity and acceleration are going to be server-side, the client isn't going to need those notions - it'll just get new positions every however-frequently from the server

nickwn:
Quick update for everyone out of the loop: I removed the update() methods on the entity and components because:
1.  Rust was making it difficult for components to modify each other (ex. camera component modifying its entity's transform component) and 
2. It would make things too convoluted in some cases (ex. originally, each entity would be responsible for updating their own transform, but in actuality they're all just received together in a big array from the server so it would be easier to just loop over the entities and set their transforms in the game loop once they are received). 

This makes the design more similar to a data-oriented ECS design so it's really tempting just to just drop everything and do specs, but this design is a little more beginner friendly so I'll keep it mostly the same for now. The hierarchy has been giving some issues though. For example, if I wanted to find the camera in the scene every frame I would have to traverse the graph every time, (Rust wouldn't allow caching a reference to it) but if I used specs I could just cache the index since the scene is stored in a big linear array.

I also integrated the scene representation into the application, so check that out too. I also added a camera component but it's not integrated since input isn't merged yet.

chariot-client/src/application.rs
```diff
@@ -10,36 +10,76 @@ use std::{
...
 }
 
 impl Application {
     pub fn new(renderer: Renderer) -> Self {
+        let mut resources = ResourceManager::new();
+
+        let import_result = resources.import_gltf(&renderer, "models/DamagedHelmet.glb");
+
+        if !import_result.is_ok() {
+            panic!("Failed to import model");
```
KNOXDEV, line 28: This panics for me currently since `models/DamagedHelmet.glb` is not in the repo. I changed it to `models/FlightHelmet/FlightHelmet.gltf` but once it loaded, I still only got a white screen. Is this code supposed to render at the moment?

KNOXDEV (changes requested):
Just one thing I want addressed before merging in. 
I don't pretend to fully understand the rest of the code right now, but I do stand by my original comment that you've reiterated in the review that I think a simpler model than a full ECS would be easier to work with.
Overall, still good work.

KNOXDEV:
Read your comment on Discord and answered my question for myself. Still not sure I want panicking code merged into main, so I'll approve this for now, but I would merge in the render code first if possible.

KNOXDEV (approved):

chariot-client/src/application.rs
```diff
@@ -18,11 +20,54 @@ pub struct Application {
...
+
+        let mut world = World::new();
+        let mut helmet = Entity::new();
+        helmet.set_component(Transform {
+            translation: glam::Vec3::ZERO,
+            rotation: glam::Quat::from_axis_angle(glam::Vec3::X, f32::to_radians(90.0)),
+            scale: glam::vec3(0.3, 0.3, 0.3),
+        });
+
+        // temporarily commenting this since the new import stuff is in a different branch
```
trulyronak, line 62:
```suggestion
```
don't need this right

trulyronak (approved):
🚀

PR #17, opened 2022-04-14
Adding WebSocket Connection for Client

trulyronak:
*heavily WIP*

Current plan:

- Use approach from https://blog.logrocket.com/how-to-build-a-websocket-server-with-rust/ (rn it's the same code, this is entirely for proof of concept testing right now, i'll be modifying / narrowing down to what's needed shortly) to setup a simultaneous web socket connection while we start the main game loop in the networking
- create some sort of shared packets / shared input/output stream that the web sockets and the tcp game loop can share / use to communicate with each other
- send messages back and forth with each other


trulyronak:
ff to leave feedback now but keep in mind this is in *no way* even ready to review yet, just wanted to get it out there in case i'm doing something eggregiously wrong atm

chariot-server/Cargo.toml
```diff
@@ -7,4 +7,10 @@ edition = "2021"
 
 [dependencies]
 glam = "0.20.3"
+tokio = { version = "0.2", features = ["macros", "sync"] }
```
KNOXDEV, line 10: I know you haven't started yet, but I fully and wholeheartedly discourage adding the tokio runtime to our project. 

Unless you have a strong reason for doing so that is. It's a large, chunky dependency and I predicted we wouldn't need an async runtime for this project. 

It's a bit like Meteor.js to host a static, single page website. 

Again, if you have a strong reason for requiring a runtime, then fine, but I don't currently know why we would.

KNOXDEV: Just looking at the websocket library for rust, its API is basically 1-1 with the TCPStream api we're currently using, so I would imagine the websocket code we would need would not be that much more extensive than what we already have for networking. We might even be able to reuse quite a bit. 

https://docs.rs/websocket/latest/websocket/

trulyronak: thanks for letting me know! will look at that to rethink how im doing this - i had added tokio primarily because it seemed needed to support async stuff, but if you don't think we need async stuff i will work on that

trulyronak: @KNOXDEV closing this PR entirely in favor of #18 , i redid it completely

PR #18, opened 2022-04-15
Adding WebSocket Connection for Client

trulyronak:
Basing it off of comments from @KNOXDEV and primarily this repo: https://github.com/snapview/tungstenite-rs

*heavily WIP*

Current plan:

- setup a simultaneous web socket connection while we start the main game loop in the networking
- create some sort of shared packets / shared input/output stream that the web sockets and the tcp game loop can share / use to communicate with each other
- send messages back and forth with each other


KNOXDEV:
I know that this is still a heavy work in progress, but I just wanted to say that this is a lot cleaner and less heavy and easier to understand. Thanks for your hard work, Ronak!

One thing I would add at this stage is that websocket messages are currently handled in separate threads. I wonder if that's necessary, since it complicates the logic with Mutexes, etc. The current TCPStream implementation manages to handle a single threaded approach, so that might be a good example to look at if a single-threaded approach is possible.

trulyronak:
ready for review now

- mirrors much of the existing tcp networking code, so i think there's probably some code reuse I could have done (i based my models off of aarons, but found i didn't need as much stuff).
- right now, the server just echos back any web socket messages for testing purposes

to test
```
cd chariot-server/
cargo run

# in another shell
cd chariot-client/
cargo run

# in another shell
cd chariot-web/
yarn install
yarn dev
```

Why Next?
- gonna deploy it with vercel, no issues with deployment
- planning on just hard coding / maybe configuring a way to connect to a specific server
- this way we can do more fun things with the web ui (easier pwa / i just like having dynamic open graph tags)

chariot-core/src/networking/ws.rs
```diff
@@ -0,0 +1,68 @@
...
+            self.socket
+                .write_message(msg)
+                .expect("should have been able to send message");
+        }
+    }
+}
+
+mod tests {
+    #[test]
+    fn test_connection() {}
```
alistairgray42, line 67: Either have an actual test here (idk to what extent you can copy the TCP connection one but if not much modification is needed that might work) or remove this

chariot-server/src/game.rs
```diff
@@ -77,6 +96,30 @@ impl GameServer {
...
+    // creates a websocket for any audience connections
+    fn acquire_any_audience_connections(&mut self) {
+        self.ws_server
+            .set_nonblocking(true)
+            .expect("non blocking should be ok");
+
+        let stream_option = self.ws_server.incoming().next();
+        match stream_option {
+            Some(stream_result) => match stream_result {
+                Ok(stream) => {
```
alistairgray42, line 108: `if let` (https://doc.rust-lang.org/rust-by-example/flow_control/if_let.html) is the more idiomatic way to do both the `Some` and the `Ok` matchings here (for when you don't want to do anything in the other `None` or `Err` cases)

chariot-server/Cargo.toml
```diff
@@ -7,4 +7,5 @@ edition = "2021"
 
 [dependencies]
 glam = "0.20.3"
+tungstenite = "0.16.0"
```
KNOXDEV, line 10: Fun trick: you can reexport tungstenite as a module from chariot-core. This is considered the idiomatic way to do it, rather then using the dependency twice.

https://github.com/rust-lang/api-guidelines/discussions/176

chariot-server/src/game.rs
```diff
@@ -85,12 +128,57 @@ impl GameServer {
                     ServerUpdatingPacket::Ping => {
                         println!("Received a Ping packet from client #{}!", i);
                         connection.push_outgoing(ClientUpdatingPacket::Pong);
+                        self.ws_connections.iter_mut().for_each(|ws| {
```
KNOXDEV, line 131: Add a comment here indicating this should be removed at some point so we don't forget haha.

chariot-server/src/game.rs
KNOXDEV: I would actually say do a while let, loop on incoming until you've depleted it.

chariot-server/src/game.rs
```diff
@@ -85,12 +128,57 @@ impl GameServer {
...
+                    tungstenite::Message::Pong(_) => {
+                        println!("got message from client #{} of type Pong", i)
+                    }
+                    tungstenite::Message::Close(_) => {
+                        println!("got message from client #{} of type Close", i)
+                    }
+                }
+            }
+        }
+        if message_to_send.len() > 0 {
```
KNOXDEV, line 175: I would create a send_ws_broadcast function instead so you can just call it from each specific match case.

KNOXDEV (changes requested):
very good, just a few minor nit picks, but no major changes at this point.

Great work, Ronak!

KNOXDEV (approved):
Great work Ronak!

PR #19, opened 2022-04-17
Crazy button stuff

TheCrossBoy:
Rust go brrrr so there's two ways you can deal w input atm:

1. Implement the trait `Watching`, with various input functions, then have the GameClient call whatever object needs to handle input. This is probably most useful for things like menu UIs, where there might be lots of buttons or things where it's easier to handle each instance on itself
2. Use GameClient.pressed_buttons, which is a set containing all pressed buttons

I'll do some work to clean it all up and make it less messy (and uncomment the stuff I commented so I could test without a server)


chariot-client/src/game.rs
```diff
@@ -1,8 +1,41 @@
+use crate::client_events::Watching;
 use chariot_core::networking::{ClientUpdatingPacket, ServerConnection, ServerUpdatingPacket};
+use std::collections::HashSet;
 use std::net::TcpStream;
+use winit::event::{ElementState, VirtualKeyCode};
 
 pub struct GameClient {
     connection: ServerConnection,
+    pressed_keys: HashSet<VirtualKeyCode>,
```
nickwn, line 9: Can we store this on the Application? Otherwise NBD

nickwn (approved):
LGTM, Just merge main

chariot-server/src/game.rs
```diff
@@ -136,6 +137,26 @@ impl GameServer {
                         //     )));
                         // })
                     }
+                    ServerUpdatingPacket::InputToggle(event, enable) => match event {
+                        InputEvent::Engine(status) => {
+                            if !enable {
+                                // self.players[figureOutWhoThisIs()].player_inputs.engine_status = EngineStatus::Neutral;
```
KNOXDEV, line 143: this will just be `self.game_state.players.get(i)` I think.

chariot-client/src/main.rs
```diff
@@ -17,25 +18,72 @@ fn main() {
...
-    // game_client.ping();
-
     let event_loop = winit::event_loop::EventLoop::new();
     let context = renderer::context::Context::new(&event_loop);
     let renderer = renderer::Renderer::new(context);
-    let mut application = application::Application::new(renderer);
+    let mut application = application::Application::new(renderer, game_client);
+
+    // Example of main loop deferring to elsewhere
+    event_loop.run(move |event, _, control_flow| {
```
KNOXDEV, line 27: This `event_loop` block is kind of sprawling and difficult to read, but I don't have any advice on how to fix it atm. Just noting.

chariot-client/src/game.rs
```diff
@@ -38,4 +40,10 @@ impl GameClient {
             }
         }
     }
+
+    pub fn send_input_event(&mut self, event: InputEvent, enable: bool) {
```
KNOXDEV, line 44: Idk about this boolean `enable`. Is this just indicating whether the input has started or stopped? I feel like that's still just an input event, but with `EngineStatus.Neutral`, etc.

KNOXDEV (changes requested):
Just minor points this time Josh. Not everything necessarily needs responding to. Great work overall!

chariot-client/src/main.rs
TheCrossBoy: I attempted a bit of something to try to fix this, not fully solved but it's at least a bit less cluttered I think

chariot-client/src/game.rs
TheCrossBoy: I originally had something typed out to explain my reasoning, but I think you're actually right that this would make more sense to figure out on the client.

I'll do some shifting of logic and such to have this determination be here instead.

TheCrossBoy: Alrighty, this has been moved to the client with the following solution:

- `Application.get_input_event` will turn a key into an input
- `Application.invert_event` will attempt to turn a given input event into its "inverse" (accel/brake -> neutral, etc), when relevant
- If an input's event has an inverse, then it will send that signal to the server. This has the added benefit of not affecting anything when we do "single click" buttons like abilities/powerups/etc, because it just won't have an inverse and thus just won't send any packets to the server when lifted.

KNOXDEV (approved):
Looks great! The extra comments are a nice touch. Thanks Josh!

PR #22, opened 2022-04-21
Third dimension c:

alistairgray42:
Steps to complete this PR:

- [x] Add a struct field to represent the upward direction for each object
- [x] Add a skeleton for finding course height and use this to compute a new upward direction each physics tick
- [x] Modify the normal force so that it takes into account the direction that objects are facing up
- [x] Modify the normal force and rolling resistance so that they only apply when objects are resting on the ground, not in flight
- [x] Modify bounding boxes to take into account rotation in the y-direction
- [x] Write tests for the brain-hurty geometry 

Things to pay attention to during review:
- Computing the three dimensions of the bounding box includes one rotation in the XZ-plane and one in a plane by a basis containing Y and the corner of XZ instead of actually doing a 3D rotation (which would be likely slower as well as very brain-hurty)
- Since we're not really sure yet how we're getting the height at any location, I'll be using a dummy function - is there anything better to do?


chariot-server/src/physics/collisions.rs
```diff
@@ -40,53 +88,52 @@ impl PlayerEntity {
...
-        let mut theta: f64 = heading.dot(zero_angle_vec).acos();
-        if theta > std::f64::consts::FRAC_PI_2 {
-            theta -= std::f64::consts::FRAC_PI_2;
-        }
+        let xz_corner = DVec2::new(x_2, z_2);
+
+        // Since the extremum distance in the y-direction is going to be the
+        // same no matter how much we spin the body about the upward direction,
+        // we can use any plane which contains the upward direction. The easy
+        // choice is to pick the plane that also goes through the origin and
```
alistairgray42, line 101: Unfortunately, while this is the easiest choice it is also not quite correct - this will often overestimate the size of the bounding box (since e.g. if the tilt direction is through the middle of an edge of the solid, the corner won't rotate as much as in this case). However, it should never underestimate the size of the bounding box, and the error shouldn't be too much for tall-ish skinny-ish objects (which is what chairs are). Unfortunately, I haven't been able to figure out how to do this with the correct basis (which is spanned by `unit_upward_direction` and the y-axis) without doing 3D rotations, which are nasty and computationally very expensive. so yeah i think this is adequate, esp at the stage of development we're at, but I wanted to flag that

KNOXDEV: I'm not going to pretend I understand all of this, but I'm fairly confident this is a sufficient approach. This seems to agree with approaches I'm familiar with in other games.

KNOXDEV (approved):
Looks good to me Alistair! Good work.

PR #24, opened 2022-04-23
Render graph take 2

nickwn:


chariot-client/src/renderer/mod.rs
```diff
@@ -335,6 +408,103 @@ impl Renderer {
...
+                });
+
+        encoder.begin_render_pass(&wgpu::RenderPassDescriptor {
+            label: None,
+            color_attachments: &color_attachments,
+            depth_stencil_attachment: depth_stencil_attachment,
+        })
+    }
+
+    fn is_same_framebuffer(next: Option<&(&str, &[RenderItem])>, fb_name: &str) -> bool {
```
KNOXDEV, line 460: The string matching here feels a little weird, especially since it's performed repeatedly in the render loop. What exactly does this function do? Is it a "hack"?

chariot-client/src/renderer/mod.rs
```diff
@@ -335,6 +408,103 @@ impl Renderer {
         });
     }
 
+    fn new_wgpu_render_pass<'a>(
```
KNOXDEV, line 411: The naming of this function is a little awkward. Is it called "new" to differentiate from the old one? If so, it should just be called `wgpu_render_pass`

KNOXDEV (changes requested):
Thanks Nick. I know this is a draft right now, but you posted it in Discord along with the other changes, so I'm taking a look at it anyways. Feel free to ignore irrelevant feedback.

Some overall points: 

* I notice a lot of unused functions and variables, particularly in `renderer/mod.rs`. This is common when old and new code share space, but just make sure you delete everything that isn't being used when you're done. It can be a bit difficult to understand what parts of the code do things.
* I also notice, particularly in `render_job.rs`, that lifetimes are used with most structs and functions. This can happen easily when you end up creating a high-level functionality that requires lifetimes: it will trickle down to all downstream things that use it since its part of the type signature. Unfortunately, its kind of difficult to follow, so if in those cases, there are ways for you to use a smart pointer type like `Rc` or `Cow`, I would just do that instead honestly.
* Overall, there's definitely several files in the graphics section that are hundreds of lines long with very few comments (`render_job.rs`, etc). I would prefer to see more comments along the way, especially since many people on our team (myself included) that do not have a graphics background and will have a harder time providing good code review otherwise.

Thanks for all the work Nick, your expertise is greatly appreciated.

chariot-client/src/renderer/mod.rs
nickwn: Yeah it is stupid, I just wanted to check if the last pass was using the same framebuffer but it's weird since at the callsite a) there might not be a next pass (finished iterating over the job) and b) passes can sometimes have no framebuffer (compute passes) so as a one-liner in the while loop conditional it was a little large.

nickwn:
Thanks for the review, I think I fixed most of the smaller stuff and cleaned up the functions I don't plan to use in the future. 

As far as lifetimes, to some extent they are hard to avoid because wgpu uses them but I admit I do still use them in a lot of areas in cases like parameter structs and other structs that are meant to be temporarily used on the stack. The big things that are meant to be stored by the application (like StaticMeshDrawable) don't use any lifetimes.

PR #25, opened 2022-04-25
just do 3d rotation instead of the overcomplicated nonsense lol

alistairgray42:
as per title

this uses glam's builtin goodness instead of trying to reinvent the wheel

verification: all the same tests pass! also this should make enough sense that a human can understand it


KNOXDEV (approved):
nice ™️

alistairgray42:
:rocket:

PR #26, opened 2022-04-25
Add list of packet types

alistairgray42:
Hopefully this helps compartmentalize our lifecycle and the things we'll need to handle (sending and receiving each of these packet types should mostly be separate tasks?)

Suggestions for anything more that should go here would be great, this is just what I whiteboarded in five minutes

Keeping this as a draft / not opening for review yet because I want to write a bit of server-side handling of some of this (sufficient to power the MVP) in this PR as well

### DIFF
--- a/chariot-core/Cargo.toml
+++ b/chariot-core/Cargo.toml
@@ -6,7 +6,7 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-glam = "0.20.3"
+glam = { version = "0.20.3", features = ["serde"] }
 config = { version = "0.13", features = ["yaml"], default-features = false }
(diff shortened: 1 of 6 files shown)


chariot-server/src/chairs.rs
```diff
@@ -0,0 +1,50 @@
+use chariot_core::{
+    entity_location::EntityLocation,
+    player_inputs::{EngineStatus, PlayerInputs, RotationStatus},
+};
+use glam::DVec3;
+
+use crate::physics::player_entity::PlayerEntity;
+
+// We could implement something to load mass and size from a file or whatever,
```
trulyronak, line 9: Would be nice to have a chairs.json file in core that stores name, stats, path to obj files - something to add later

trulyronak (approved):
🚀

PR #27, opened 2022-04-28
MVP client-server integration

alistairgray42:
What this PR does:

- [x] Add a `GraphicsManager` struct to split off much of the graphics functionality formerly in `Application`, and trims a bit off `GameClient` as well, so that `Application` can focus on business logic
- [x] Add an `EntityID` component so that specific entities (i.e. players) can be re-accessed. (tbh this exists to get around Rust's ownership rules)
- [x] Add `add_player` and `update_player_location` methods, and calls them when appropriate packets are received from the server  so that the client remains updated with the server's representation of physics
- [x] Update the server with the client's inputs


alistairgray42:
@nickwn Any idea what this might be? Happens for me on this branch and not on main (and happens before model loading)
image

nickwn:
Is it rendering a frame with no models?

alistairgray42:
you should now be able to:
- hold A/D and see the helmet s p i n
- press W and see it z o o m off into the distance

chariot-client/src/application.rs
```diff
@@ -1,185 +1,63 @@
...
     }
 
     pub fn update(&mut self) {
-        let surface_size = self.renderer.surface_size();
-        let surface_size = glam::Vec2::new(surface_size.width as f32, surface_size.height as f32);
         let mouse_pos = glam::Vec2::new(self.mouse_pos.x as f32, self.mouse_pos.y as f32);
+        self.graphics.update(mouse_pos);
 
-        let rot_range = glam::Vec2::new(std::f32::consts::PI, std::f32::consts::FRAC_PI_2);
+        self.game.sync_incoming();
```
trulyronak, line 40: nit: not a huge fan of this naming convention (@KNOXDEV can chime in)

I'd name this to be `get_incoming_packets` (since that's what this method actually does)

chariot-client/src/application.rs
```diff
@@ -1,185 +1,63 @@
...
-        dfs_mut(self.world.root_mut(), &|e| {
-            if let Some(camera) = e.get_component::<Camera>() {
-                let norm_orbit_angle = (mouse_pos / surface_size) * 2.0 - 1.0;
-                let orbit_angle = norm_orbit_angle * rot_range;
-                let new_camera = Camera {
-                    orbit_angle,
-                    ..*camera
-                };
-                e.set_component(new_camera);
+        for packet in self.game.get_incoming_packets() {
```
trulyronak, line 42: nit: in keeping with my prior comment, I'd name this simply `incoming_packets()` or `get_current_packets`

chariot-client/src/application.rs
```diff
@@ -1,185 +1,63 @@
...
+            match packet {
+                ClientBoundPacket::Pong => {
+                    println!("Received a Pong packet from server!");
+                }
+                ClientBoundPacket::Message(text) => {
```
trulyronak, line 47: nit:  i think we can probably now remove this and the pong data types - not needed anymore since we've added real stuff now

trulyronak (approved):
lgtm

PR #28, opened 2022-04-28
Add models

nickwn:


chariot-client/src/application.rs
```diff
@@ -49,24 +45,50 @@ impl Application {
...
 
-        world.root_mut().add_child(helmet);
+        {
+            let chair_import_result = resources.import_gltf(&mut renderer, "models/chair.glb");
+
+            let mut chair = Entity::new();
+            chair.set_component(Transform {
+                translation: glam::vec3(0.0, 0.5, 0.0),
+                rotation: glam::Quat::IDENTITY,
+                scale: glam::vec3(1.1995562314987183, 2.2936718463897705, 1.1995562314987183) * 0.2,
```
alistairgray42, line 57: ? where do these numbers come from

alistairgray42 (approved):
Looks good to me. :rocket: 

The camera is also kind of close to the model? Don't think that's much of a concern at the moment

chariot-client/src/application.rs
nickwn: the chair had some weird scaling which wasn't read in when importing the gltf file so I just manually copied and pasted the scaling here.

PR #30, opened 2022-05-01
Triggers & Checkpoints

TheCrossBoy:
Added a system for triggers:

- Triggers are basic entities that have a static predefined position
- Colliding with them does _something_
- Added three triggers:
    - MinorCheckpoint
    - MajorCheckpoint
    - FinishLine

The logic for these follows the MarioKart Wii logic (with one change to fix a bug they had xd):

- Across the map, we define boxes for minor checkpoints, major checkpoints, and the finish line
- Minor checkpoints are exclusively used to reset your position if you fall off the map (we could also use it for ais.... if we got that far lol), if we don't have time we can just ignore them completely
- Major checkpoints each have a unique, constantly increasing zone identifier as you go through the map (the first one in front of the finish line is 1)
- Crossing a major checkpoint trigger will increase the player's zone ID by 1, but _only_ if their zone id is equal to the previous one 
- Crossing the finish line will set your zone ID to 0 and increase your lap count, but _only_ if your zone ID is the last one in the map

In effect, this means that you must cross each major checkpoint in order, then the finish line, in order for a lap to count. This solves the problem of cheating, assuming that we properly setup good major checkpoints across a map (which hypothetically shouldn't be that hard). This should also only actually let someone get a single lap, because their zone will be updated to 0, preventing the following ticks from triggering extra laps.

This still allows for shortcuts in maps, as we just have to make sure the major checkpoints are placed in a smart way such that you will always go through them (or extend them so shortcuts also pass through them).

Oh as a bonus I just set something up to contain the information about maps, but nothing really fully implemented on that yet, mostly just temporary until we have a real solution later.

As of now, all the code side of this should be implemented, I'd just like to poke around with blender a bit to see if this type of stuff can be coded directly into the models, so we don't have to painstakingly put them in with code. As such, I'll also have to play around with how that stuff loads in to make sure it loads in the correct way (and doesn't try to do collisions with players).

Please let me know if you have any questions/concerns/comments!


chariot-server/src/game.rs
```diff
@@ -10,14 +10,17 @@ use chariot_core::player_inputs::InputEvent;
...
 use crate::physics::player_entity::PlayerEntity;
+use crate::physics::trigger_entity::TriggerEntity;
 
 pub struct GameServer {
     listener: TcpListener,
     ws_server: TcpListener,
     connections: Vec<ClientConnection>,
     ws_connections: Vec<WebSocketConnection>,
     game_state: ServerGameState,
+    map: Option<Map>,
```
alistairgray42, line 23: I think this is fine as an Option for now, but in the long term we should try to discourage making this nullable (i.e. switch this over once we have a map implemented), because there's no need for this to be nullable (we'll have at least one map, and one of them is going to be the "default")

chariot-server/src/physics/trigger_entity.rs
```diff
@@ -0,0 +1,30 @@
+use crate::physics::player_entity::PlayerEntity;
+use glam::DVec3;
+
+type BoundingBoxDimensions = [[f64; 2]; 3];
+
+pub trait TriggerEntity {
+    fn get_bounding_box(&self) -> BoundingBoxDimensions;
+
+    fn check_bounding_box_collisions(&self, ply: &PlayerEntity) -> bool {
```
alistairgray42, line 9: nit: i don't think `ply` is a recognizable enough abbreviation to be useful

chariot-server/src/physics/trigger_entity.rs
```diff
@@ -0,0 +1,30 @@
+use crate::physics::player_entity::PlayerEntity;
+use glam::DVec3;
+
+type BoundingBoxDimensions = [[f64; 2]; 3];
+
+pub trait TriggerEntity {
```
alistairgray42, line 6: Code duplication in here is why I was thinking of making this a trait that also encompasses players as well as checkpoints and powerups - players could have a `trigger` method that applies the velocity change from a collision, and we'd have to treat them differently in business logic, but physics-wise what they're doing is pretty much exactly the same.

chariot-client/src/game.rs
```diff
@@ -1,10 +1,12 @@
+use chariot_core::lap_info::LapInformation;
 use chariot_core::networking::{ClientBoundPacket, ServerBoundPacket, ServerConnection};
 use chariot_core::player_inputs::InputEvent;
 use std::net::TcpStream;
 use winit::event::{ElementState, VirtualKeyCode};
 
 pub struct GameClient {
     connection: ServerConnection,
+    lap_info: LapInformation,
```
alistairgray42, line 9: I think this belongs in `Application`, not `GameClient` (which I think should be pretty much just an implementation of a networking API)

alistairgray42: And you aren't using this anywhere on the client yet, but I don't think the client will need to know this - the `Application` will be telling the `GameClient` to send lap update packets, so having an array of this in `Application` makes more sense to me

chariot-server/src/physics/mod.rs
```diff
@@ -77,10 +80,17 @@ impl PlayerEntity {
             size: self.size,
             bounding_box: self.bounding_box,
             physics_changes: self.physics_changes.clone(),
+            lap_info: self.lap_info,
         };
 
         new_player.apply_physics_changes();
 
+        for b in potential_triggers.iter() {
```
alistairgray42, line 88: This should go before the `apply_physics_changes` method call at least, so that if the `trigger` method adds a physics change from a powerup, it gets the chance to apply one tick more snappily.

chariot-server/src/game.rs
```diff
@@ -242,3 +260,9 @@ impl GameServer {
     // queue up sending updated game state
     fn sync_state(&mut self) {}
 }
+
+pub struct Map {
```
alistairgray42, line 264: is this supposed to be in `map.rs`?

chariot-server/src/game.rs
```diff
@@ -223,6 +227,20 @@ impl GameServer {
             player.set_upward_direction_from_bounding_box();
         }
 
+        let triggers: &mut Vec<Box<&dyn TriggerEntity>> = &mut Vec::new();
```
alistairgray42, line 230: I think this doesn't need to run each time `simulate_game` does. I'd add a method to setup the map and store the list of triggers in the server state so that we don't need to construct the whole list every time. (flagging also that, this way, there isn't an incredibly intuitive way to "expire" powerups so that they can only be picked up after however-long a period of time - with a list stored in server state, we can add a pickup re-allowed time in there.)

chariot-server/src/checkpoints.rs
```diff
@@ -0,0 +1,96 @@
...
+    pub bounding_box: BoundingBoxDimensions,
+}
+
+impl MajorCheckpoint {
+    pub fn new(id: MajorCheckpointID, pos: DVec3, size: DVec3) -> Self {
+        Self {
+            id: id,
+            pos: pos,
+            size: size,
+            bounding_box: [
```
alistairgray42, line 39: This only allows major checkpoints that are aligned to the XYZ basis, and I think that's fine.

chariot-server/src/physics/trigger_entity.rs
alistairgray42: Hmm, this might lead to some issues that would be annoying to deal with (e.g. mutable state is the easy way to deal with triggers but collisions can't do that because of order concerns) - totally understandable why you've done it this way.

alistairgray42 (changes requested):
I like this PR! Lots of good stuff going on here. 
I want to see some unit tests (logic is similar to the collision tests, so you could model any tests off of those) before approving; I would say the feedback about merging the TriggerEntity trait and the behavior of players into a single trait is non-blocking, but definitely something to consider at least.

chariot-server/src/game.rs
TheCrossBoy: For sure, this was definitely very temporary. 

It was also because before we implement a state of "menu" vs "in game", there may not be a map yet (especially if we add more than one track).

chariot-server/src/physics/trigger_entity.rs
TheCrossBoy: Ah, the game I used to code in used ply for everything so it was entirely habit. I'll update!

chariot-server/src/physics/mod.rs
TheCrossBoy: I think my main concern with this was that you won't actually be considered as having hit the trigger until the tick *after* you're inside of it, which might be slightly more disorienting. 

We also need to account in some way for the chance that you move entirely through the trigger on a tick, but that is fairly unlikely if we make them large enough so we can address that later hopefully.

chariot-server/src/game.rs
TheCrossBoy: Agreed, this was only a temporary solution since they're empty anyways. 

This next week I'm going to work on getting the track file to load this stuff, which is where this will be actually calculated.

chariot-server/src/checkpoints.rs
TheCrossBoy: Yeah it should be fine since that's mostly an us thing anyways. If it becomes a limiting problem, I'll fix that!

TheCrossBoy:
From what we decided today: this is now just getting the framework merged in, and we will handle the whole "server opening maps" thing in a future PR (to be created in a bit), which is where we'd actually load this stuff in.

alistairgray42 (approved):

PR #31, opened 2022-05-06
Tweak physics

alistairgray42:
quick little PR which:
- Adds maximum velocity and angular velocity parametrized constants so that we don't have hilariously unlimited vroom vrooms
- Messes with parameters to make driving around on our track feel possible + better

Steps for review:
- [ ] literally anyone, just pull and run and see how the physics feels - feel free to suggest / directly make commits changing the parameters

the model i'm working off of is that physics should feel the following ways:
- like, kinda awkward? like, none of us will have quite gotten the hang of it and we'll be doing nonoptimal routes and running into each other a lot
- not incredibly fast but top speed is a little too fast to be comfortable
- "stop accelerating, turn 90 degrees, and start accelerating to basically drift" is a viable strat

these will not be final values but yeah

### DIFF
--- a/chariot-core/src/settings.rs
+++ b/chariot-core/src/settings.rs
@@ -45,22 +45,24 @@ impl Settings {
             // smaller (~30 times smaller is realistic) than
             // `rolling_resistance_coefficient` to have the (correct) property
             // that drag dominates at higher speeds.
-            .set_default("drag_coefficient", 0.01)?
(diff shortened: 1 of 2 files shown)


KNOXDEV (approved):
Wow this is one hell of a way to jump back into this project

I cannot tell what is happening right now but I think it's a good speed and drift so far.

I say merge ASAP so others can use this. We'll be able to refine further once we have...a cohesive reality

PR #32, opened 2022-05-07
Audience Interaction

trulyronak:
This adds the ability to send out question prompts to all connected web clients to vote on what happens.

### How to test
- do `yarn dev` in the web folder to get the web client running, then go to localhost:3000 to see the app
- do `cargo run` in server and client to get the server running
- join active game via web ui
- vote for something
- wait 30 seconds
- see if your vote won!

### Things to note
- i think I created alot of different structures for this, would appreciate help in terms of how i might simplify stuff
- how would the methods i created flow with the server code / make it easy to send requests
- how i handled serialization; i used serde_json to make communicating with the web easier, but I wasn't able to get the same `trait` style implementation working that @KNOXDEV got working really nicely (would appreciate feedback / code on how to do that if you think it's nicer)
- Netlify builds aren't working yet, but that's because I'm tweaking stuff (vercel is nicer, but vercel doesn't have free deployment of organizations so i'm using netlify)


chariot-core/src/networking/packets.rs
```diff
@@ -47,3 +49,19 @@ pub trait Packet: Serialize + DeserializeOwned {
...
+#[derive(Serialize, Deserialize, Clone)]
+pub enum WSAudienceBoundMessage {
+    Prompt(Prompt), // Question, 4 Answer Choices
+
+    Winner(i32), // The winning choice (tuple index)
+
+    Assignment(Uuid), // Sends a uuid that the server will use to identify the client
+}
+
+pub type Prompt = (String, (String, String, String, String));
```
KNOXDEV, line 62: Maybe this should be called something different to distinguish it from the `WSAudienceBoundMessage::Prompt` type?

Also, do we want to include a number of seconds to count down the vote?

chariot-core/src/networking/ws.rs
```diff
@@ -37,27 +40,45 @@ impl WSConnection {
...
                 if msg.is_binary() || msg.is_text() {
                     // this is where we handle shit
-                    self.incoming_packets.push_back(msg);
+                    let txt = msg
+                        .to_text()
+                        .expect("should have been able to convert message to string");
+                    let message_result: Result<WSServerBoundMessage, Error> =
+                        serde_json::from_str(txt);
+                    if message_result.is_err() {
+                        // self.incoming_packets.push_back(msg);
```
KNOXDEV, line 49: maybe print the error here? Probably shouldn't panic because then web clients can take down the server 😲

chariot-core/src/networking/ws.rs
```diff
@@ -37,27 +40,45 @@ impl WSConnection {
             Ok(msg) => {
                 if msg.is_binary() || msg.is_text() {
```
KNOXDEV, line 41: it looks like we only handle the case where the message is text, so maybe leave out the is_binary check?

chariot-server/src/game.rs
```diff
@@ -52,6 +78,8 @@ impl GameServer {
     pub fn start_loop(&mut self) {
         let max_server_tick_duration = Duration::from_millis(GLOBAL_CONFIG.server_tick_ms);
 
+        let mut runOnce = true;
```
KNOXDEV, line 81: code smell check

I understand this is for verifying the behavior of the interaction system, but this needs to be removed or moved into a test module before merging with main IMO

chariot-server/src/game.rs
```diff
@@ -151,48 +225,47 @@ impl GameServer {
...
-                    }
-                    Message::Ping(_) => {
-                        println!("got message from client #{} of type Ping", i)
-                    }
-                    Message::Pong(_) => {
-                        println!("got message from client #{} of type Pong", i)
-                    }
-                    Message::Close(_) => {
-                        println!("got message from client #{} of type Close", i)
+                    chariot_core::networking::WSServerBoundMessage::Vote(id, vote) => {
```
KNOXDEV, line 231: I think you should use WSServerBoundMessage to make the :: a little less nested here

chariot-server/src/physics/mod.rs
```diff
@@ -103,18 +103,18 @@ impl PlayerEntity {
...
-            self.gravitational_force_on_object(),
-            self.air_resistance_force_on_object(),
-        );
-        println!(
-            "is aerial? {} if not: normal {}, rolling resistance {}",
-            self.is_aerial(),
-            self.normal_force_on_object(),
-            self.rolling_resistance_force_on_object()
-        );
+        // println!(
```
KNOXDEV, line 115: feel free to just remove this for now

chariot-web/package.json
```diff
@@ -14,7 +14,8 @@
 	"dependencies": {
 		"next": "12.1.5",
 		"react": "18.0.0",
-		"react-dom": "18.0.0"
+		"react-dom": "18.0.0",
```
KNOXDEV, line 17: REACT GAME

KNOXDEV (changes requested):
Some minor points. Great work Ronak!

You'll notice I did not comment on the web side of things really. That's partially because its fine and partially because we don't really need to spend any time making sure it maintainable by the entire team.

chariot-core/src/networking/packets.rs
trulyronak: choosing not to include number of seconds here, but added it to the config file

KNOXDEV (approved):

PR #38, opened 2022-05-10
State management

alistairgray42:
- Makes server-side state more explicit by introducing a `GamePhase` enum, which denotes when certain kinds of packets are expected/allowed/handleable
- Moves some code out of `server/game.rs`, to better break things up
- Handles server-side logic as to when to transition between phases
- Notifies the clients of phase transitions as they occur
- Adds a bypass config file option to ignore all this and let client folks not have to worry about rounding up three friends to play with when developing (default on, so end-user behavior is unchanged as is

would like comments on the names of the phases, naming things is hard


chariot-server/src/game/mod.rs
```diff
@@ -0,0 +1,276 @@
...
+                .for_each(|con| con.fetch_incoming_packets());
+
+            // poll for ws input events
+            self.ws_connections
+                .iter_mut()
+                .for_each(|(_, con)| con.fetch_incoming_packets());
+
+            self.process_incoming_packets();
+            self.process_ws_packets();
+            self.check_audience_voting();
```
trulyronak, line 87: this should go inside `simulate_game`

chariot-server/src/game/mod.rs
```diff
@@ -0,0 +1,276 @@
...
+    map: Option<Map>,
+}
+
+pub struct ServerGameState {
+    phase: GamePhase,
+
+    // gets its own slot because it persists across several phases; is awkward to behave identically in all
+    players: [PlayerEntity; 4],
+
+    waiting_for_player_ready_state: WaitingForPlayerReadyState,
```
TheCrossBoy, line 39: Is there a reason why these are all being stored separately, rather than having them be part of the GamePhase enum? i.e. something like:
```rust
enum GamePhase {
    WaitingForPlayerReady(WaitingForPlayerReadyState),
    // etc
}
```

I don't think we'd need to keep the CountingDownToGameStartState stored around within the server's state when we are, say, in the PlayingBeforeVotingState, right?

chariot-client/src/graphics.rs
```diff
@@ -35,7 +35,7 @@ fn setup_world(resources: &mut ResourceManager, renderer: &mut Renderer) -> (Wor
     world.register::<Light>();
     let world_root = world.root();
     let chair = {
-        let chair_import_result = resources.import_gltf(renderer, "models/chair.glb");
+        let chair_import_result = resources.import_gltf(renderer, "models/defaultchair.glb");
         let chair_import = chair_import_result.expect("Failed to import chair");
```
TheCrossBoy, line 39: nit: this would probably be neater as a one-liner (resources.import().expect()) since we don't care about the result otherwise 
edit: along with the other place in our codebase as a whole we do that

chariot-server/config.yaml
```diff
@@ -0,0 +1 @@
+bypass_multiplayer_requirement: true
```
TheCrossBoy, line 1: what's this file for in comparison to the settings.rs file's stuff? is this used somewhere atm?

chariot-server/src/game/phase.rs
```diff
@@ -0,0 +1,101 @@
...
+ *
+ * 3. GamePhase::PlayingBeforeVoting
+ *  The idea we had was to have idk 30 seconds at the start of the race
+ *  before voting starts; players will be driving around normally and the
+ *  clients will be oblivious to the difference between Phase 3 and 4. (client
+ *  will time itself internally)
+ *
+ * 4. GamePhase::PlayingWithVoting
+ *  All the good stuff happening at once! >:))) The transition from Phase 4 to
+ *  Phase 5 is marked by the server's AllDone packet.
```
trulyronak, line 34: I think we should have phase 'PlayingWhileVoting` and `PlayingWithDecision`, where the first stage is while players play with no effect and the second stage is players playing with the effect added. These two phases don't make enough sense to be separated to me + i feel are a bit confusing. 

```suggestion
 * 3. GamePhase::PlayingWhileVoting
 *     Players will be driving around normally and the clients will simply 
 *     display "The Audience is Deciding Your Fate" (for juice). 
 *     State: `timeUntilVotingTally`, `questionPrompt`
 *
 * 4. GamePhase::PlayingWithDecision
 *     Players will be doing their thing WITH some new effect in place. 
  *    This effect only exists during this phase. 
 *     State: `timeUntilDecisionEnds`, `winningEffect` (this would be 
  *    an enum or something so we can 
 *     (a) display it on clients and (b) do special server logic)
 *     The transition from Phase 4 to Phase 5 is marked by the server's
 *     AllDone packet (sent when someone wins the game)
```

chariot-server/src/game/mod.rs
```diff
@@ -0,0 +1,276 @@
...
+                        .countdown_end_time
+                {
+                    let time_until_voting_enabled = Duration::new(30, 0);
+                    self.game_state
+                        .playing_before_voting_state
+                        .voting_start_time = now + time_until_voting_enabled;
+                    self.game_state.phase = GamePhase::PlayingBeforeVoting;
+                }
+            }
+            GamePhase::PlayingBeforeVoting | GamePhase::PlayingWithVoting => {
```
trulyronak, line 200: would be nice to have some code that runs for both stages + some code that only runs for one stage (so it can do mutual code + phase specific code)

chariot-server/src/game/mod.rs
```diff
@@ -0,0 +1,276 @@
...
+                    connection.push_outgoing(ClientBoundPacket::LocationUpdate(locations));
+                }
+            }
+
+            GamePhase::AllPlayersDone => todo!(),
+        }
+    }
+}
+
+pub struct Map {
```
TheCrossBoy, line 272: mildly annoyed that i've moved this to a seperate file 3 times but because git is funny it somehow never actually happened in the mess of a history i was dealing with but alas

so i now return this suggestion to you: probably smart to move this to its own file within the game module, since it'll get expanded upon

chariot-server/src/game/phase.rs
```diff
@@ -0,0 +1,101 @@
...
+ *
+ * 5. GamePhase::AllPlayersDone
+ *  Show standings, perhaps a retry button, any other end-of-race stuff
+ */
+
+pub enum GamePhase {
```
TheCrossBoy, line 40: See earlier comment about storing states within these

chariot-server/src/game/voting.rs
```diff
@@ -0,0 +1,115 @@
...
+
+    // prompts for a question
+    pub fn start_audience_voting(
+        &mut self,
+        question: String,
+        option1: String,
+        option2: String,
+        option3: String,
+        option4: String,
+        poll_time: Duration,
```
TheCrossBoy, line 100: is it possible to have this be an arbitrary number of options (in a vec) instead of 4? i.e. question: String, options: Vec<String>, poll_time: Duration?

I think it should be possible to send that to the websocket as well by explicitly stating the number of options, then reading that many, right?

TheCrossBoy: This is probably not something added in this PR but this is when I'm seeing it most recently, so

TheCrossBoy (changes requested):
Cool stuff, very nice organizational changes too will make it easier to find stuff going forward!

chariot-server/src/game/voting.rs
trulyronak: I'm going with 4 hard coded options just because I don't want the UI to render more than 4 options → we could do an array of fixed size 4 (which is what it is on typescript actually) but idk how force array size with rust types

KNOXDEV: `options: [String; 4]`

chariot-server/src/game/mod.rs
KNOXDEV: :+1: to this suggestion

trulyronak:
closing because i've taken over this PR and I don't want to be able to approve my own PR

PR #39, opened 2022-05-10
Peter/audio

potor10:
# Things to note

### Generic Sink implementation in AudioThread
Currently our AudioThread stores `sink: SpatialSink,` which is a https://docs.rs/rodio/latest/rodio/struct.SpatialSink.html
That being said, for playing non spatial audio, a Spatial Sink is not ideal.
This has led me to write sub-optimal code to "simulate" a non spatial sound within a spatial engine. With this code:
```
emitter_pos: [0.0; 3],
left_ear: [-0.000001, 0.0, 0.0],
right_ear: [0.000001, 0.0, 0.0]
```
Instead we want to use a Sink https://docs.rs/rodio/latest/rodio/struct.Sink.html

Is there any way to create a generic "Sink<T>" struct which can take in both SpatialSink and Sink as an input within rust?
Although both SpatialSink and Sink implement the same methods, it doesn't seem like they interface off of a trait 

image
Source code: https://docs.rs/rodio/latest/src/rodio/sink.rs.html#13-21

That way we can reserve the spatial sink for spatial playback, with a normal sink for normal playback.

### Manipulating singular audio threads
The current implementation stores all the AudioThreads in a Vec as `threads: Vec<AudioThread>` which is controlled by the AudioSource struct. This AudioSource struct is able to apply general transformations across all threads that it manages, such as volume control, pause / resume / stop, and a crossfade play (fade out all currently playing threads and then fade in a new track to play).

The issue is since the threads requires this exact AudioThread struct as an input, we can't return any specific AudioThread for more customized implementations. This comes into play when we, for example, need to continue looping a SFX (ie: a rocket thruster sound) that needs to continue looping UNTIL the game state changes (w key released). 

In this case, we shouldn't keep playing a new sound effect each time. We should instead play a sound effect once, set the infinite repeat to `True`, and monitor the specific thread individually until our game state changes. 

I wrote some code for this in Line 255:
```
// Play audio, & return a thread for self management
pub fn play_alone(&mut self, track_id: usize, ctx: &AudioCtx, opt: SourceOptions) -> AudioThread {
  let path_buf = self.tracks.get(track_id).unwrap();
  let mut thread = AudioThread::new(ctx, path_buf.to_path_buf(), opt);
  thread.set_volume(self.volume);
  thread.set_pitch(self.pitch);

  thread.play();
  return thread;
}
```
But since we can only use thread once, I can't do `self.threads.push(thread);` if I want to return the thread.
This means calling `play_alone()` will spawn a thread that is not influenced by transformations requested by the audio manager.
But it does allow the user to run their own transformations specifically on the thread.

Looking for advice on how to approach this.

### Poor performance
Poor performance when spawning many threads of audio to play at once (usually happens after 3+ Sinks simultaneously 
playing audio)

image
image

### Reading from path buffer rather than loading a source from memory
Line 117:
```
let buf = BufReader::new(self.file.try_clone().unwrap());
let source = Decoder::new(buf).unwrap();
```
Line 203: 
```
let paths = fs::read_dir(format!("./{}", path)).unwrap();

for path in paths {
  let path_buf = path.unwrap().path();
  let path_str_dsp = path_buf.display().to_string();

  let file = fs::File::open(path_buf.clone()).unwrap();

  let buf = BufReader::new(file);
  let decode = Decoder::new(buf);

  if decode.is_ok() {
    tracks.push(path_buf);
    println!("Loaded Track {}: [{}]", tracks.len(), path_str_dsp);
  }
```
Currently a new BufReader is spawned every time an audio file is played. This is done because the Vec that stores our preloaded tracks needs instances which can be reused (in multiple threads), and PathBuf is clonable. Meanwhile source is not clonable, but if we could find a way to clone it it would save us the time needed to regenerate a new source from a file path every time we request an audio playback. 

Source struct: https://docs.rs/rodio/latest/rodio/source/trait.Source.html


KNOXDEV:
Let me start by giving some high level feedback.

First, great work Peter! This is incredible stuff. The documentation is especially appreciated. 

Can you create a subfolder for the audio submodule and break up the `audio.rs` file into a file for each major struct? Its a bit easier to maintain then one giant 300+ line rust file. You can glue it all together with `audio/mod.rs`. You can also put the audio `README` into that folder, I think that would be a good place for it. https://doc.rust-lang.org/rust-by-example/mod/split.html

To address the points you've brought up in the PR:

### Generic Sink implementation in AudioThread

I think the "rusty" way to handle this is to have a AudioSink sync enum that can have either Sink or SpatialSink. When you go to use this, you'll use a match statement to disambiguate.

### Manipulating singular audio threads

Since AudioSource can't give up ownership to the AudioThreads it owns, really the easiest, if most verbose solution is to simply provide functions on AudioSource that allow you to tweak parameters for a particular thread. It's up to you if this is more preferred to what you have right now.

One other thing I'll note is that it's not a huge issue either way. Its possible we won't run into the use case at all, or if we do, rather than allow control over the AudioThread, we'll expose the exact functionality we need in that moment on AudioSource and we'll be good to go.

### Poor performance

If we run into the issue "in production" we can address it then. Not sure if we'll need more than 2-3.

### Reading from path buffer rather than loading a source from memory

I think the solution here will be to (lazily) load the file into memory and just borrow immutable references to it whenever we create a new source. The wrapping source around the buffer will copy the individual bytes as the file is played.

## some last points

Overall, I notice there's a lot of `unwrap()` since a lot of the device interaction calls could fail at any time. It would be great if instead of `unwrap()` (which crashes the program if, say, I unplug my headphones) we just kill that audio thread or don't play the sound in that moment.

In the (hopefully few) cases where we need to unwrap, it would be great if we used `expect()` and left a message that informed us of the context. 

Thats all for now Peter, and thanks again for all the work!

KNOXDEV (changes requested):
see above

chariot-client/src/audio.rs
```diff
@@ -0,0 +1,342 @@
...
+      skip_duration: Duration::ZERO,
+      take_duration: Duration::from_secs(3600),
+      pitch: 1.0,
+      emitter_pos: [0.0; 3],
+      left_ear: [-0.000001, 0.0, 0.0],
+      right_ear: [0.000001, 0.0, 0.0]
+    }
+  }
+
+  pub fn set_fade_in(&mut self, duration: Duration) {
```
alistairgray42, line 54: In general, getters and setters will tend to be less rusty than making the fields of the struct public (for non-computed properties at least) - it also shouldn't really be the case that these methods are getting called by enough other code (because of the layers on top) for this to be worth it, imo

potor10:
Hi all, checking in with some updates regarding the Audio Engine,

## Generic Sink implementation in AudioThread
Thanks to the suggestion of @KNOXDEV, it works perfectly. The user can now specify if they want a standard "sink" or a "spatial sink" depending on the values set by by the emitter position, left ear position, and right ear position in `SourceOptions`. Having a nonzero value in any of the three will enforce usage of a SpatialSink. A zero value in all 3 parameters (by default in SourceOptions) signals the usage of a standard Sink

## Manipulating singular audio threads
Since this is possibly not going to be a big issue until later, I'm putting this off for now. When we need the functionality we can choose to add more to the AudioSource impl

## Reading from path buffer rather than loading a source from memory
Also Fixed, I did some of my own research on the topic, and found that you can clone a Buffered `Source` from Rodio. Thus, I changed my implementation to utilize these Buffered sources in order to improve playback speeds.

Under suggestion of the team, I have also downloaded and uploaded all the sample audio files we have so far (unedited, with credits), and modified the readme with a basic example of usage.

potor10:
Added proper error handling instead of `unwrap()`

chariot-client/src/audio/mod.rs
```diff
@@ -0,0 +1,254 @@
...
+          i += 1;
+        }
+      };
+    }
+  }
+
+  // Play Audio
+  pub fn play(&mut self, track_id: usize, ctx: &AudioCtx, opt: SourceOptions) {
+    // Clean Up All Stopped Threads
+    self.clean();
```
KNOXDEV, line 97: running clean everywhere to prune unused audio sinks feels a little bit of a code smell to me, but after reading the rodio documentation I have yet to come up with a better solution. Either way, it'll work for now.

chariot-client/src/audio/mod.rs
```diff
@@ -0,0 +1,254 @@
...
+
+    let source = self.tracks.get(track_id);
+    let source = match source {
```
KNOXDEV, line 100: In the places where you do `match()` to handle `Some` and `None`, you might be able to use `if let` to save yourself some typing. https://doc.rust-lang.org/rust-by-example/flow_control/if_let.html

chariot-client/src/audio/mod.rs
KNOXDEV: I think the only other way to handle this is to put some kind of artificial limit on the number of audio threads and just reuse them. That said I think I prefer the way you've handled it.

chariot-client/src/audio/thread/mod.rs
```diff
@@ -0,0 +1,214 @@
...
+  // Fade out the sink
+  pub fn fade_out(self, duration: Duration) {
+    // 1 step every 10 ms
+    let steps = duration.as_millis() / 10;
+    let delta_time = duration.as_millis() / steps;
+    
+    let x = Arc::new(Mutex::new(self));
+    let alias = x.clone();
+
+    thread::spawn(move || {
```
KNOXDEV, line 131: This is a wee bit hacky. The good news is that I found how to do this in rodio! It's buried very far, I found it by using the search bar in the documentation.

I think you can do `source.take_duration().set_filter_fadeout()` https://docs.rs/rodio/latest/rodio/source/struct.TakeDuration.html#method.set_filter_fadeout

Let me know if this saves you some trouble in this method.

chariot-client/src/audio/mod.rs
```diff
@@ -0,0 +1,254 @@
...
+
+    thread.play();
+    self.threads.push(thread);
+  }
+
+  // Do a fade-out on all currently playing threads in this manager
+  pub fn fade_all_threads(&mut self, duration: Duration) {
+    // Fade Out All Currently Active Threads
+    while self.threads.len() > 0 {
+      let active_thread = self.threads.pop();
```
KNOXDEV, line 168: Just be aware that according to [Rodio](https://docs.rs/rodio/latest/rodio/struct.Sink.html), as soon as you "drop" a Sink (ie, the Sink's go out of scope), audio playback will stop. That doesn't appear to affect this particular implementation because you do some thread stuff to make the sink fadeout work, but it might come up if you change it as I've commented.

chariot-client/src/audio/mod.rs
```diff
@@ -0,0 +1,254 @@
...
+          println!("There was an issue fading out the thread");
+        }
+      }
+    }
+  }
+
+  // Pause All Audio Threads
+  pub fn pause(&mut self) {
+    let mut i = 0;
+    while i < self.threads.len() {
```
KNOXDEV, line 181: for all of the following methods that loop over the threads and call a function, I recommend you just do `self.threads.iter_mut().for_each` and pass a closure instead. http://web.mit.edu/rust-lang_v1.25/arch/amd64_ubuntu1404/share/doc/rust/html/book/first-edition/iterators.html

I won't duplicate this comment on `resume`, `stop`, `set_volume`, etc, but the same principle applies.

chariot-client/src/audio/thread/context/mod.rs
```diff
@@ -0,0 +1,25 @@
...
+pub struct AudioCtx {
+  pub _stream: OutputStream,
+  pub stream_handle: OutputStreamHandle
+}
+
+impl AudioCtx {
+  pub fn new() -> Self {
+    // Get a output stream handle to the default physical sound device
+    let output_stream = OutputStream::try_default();
+    let (_stream, stream_handle) = match output_stream {
```
KNOXDEV, line 13: in this case `expect` is equivalent.

chariot-client/src/audio/thread/mod.rs
```diff
@@ -0,0 +1,214 @@
...
+
+use std::time::{SystemTime, Duration};
+
+pub mod context;
+pub mod options;
+
+use context::AudioCtx;
+use options::SourceOptions;
+
+enum AudioSinkType {
```
KNOXDEV, line 17: Perfect!

chariot-client/src/audio/thread/mod.rs
```diff
@@ -0,0 +1,214 @@
...
+        AudioSinkType::Standard(sink) => sink.stop(),
+      }
+      // mutref.sink.stop();
+    });
+  }
+
+  // Pause playback
+  pub fn pause(&mut self) {
+    match &self.sink {
+      AudioSinkType::Spatial(sink) => sink.pause(),
```
KNOXDEV, line 165: having to do this `match` every time feels a little silly but I'm not exactly sure how we would simplify...

I think at the end of the day we have to blame the authors of Rodio for not using a trait for these structs when it was clearly the correct move.

KNOXDEV: Especially since the solution would pretty much just be wrapping these structs in another common trait.

KNOXDEV: no tasking here btw just commenting

KNOXDEV (changes requested):
Hey Peter, excellent, amazing work on the high level feedback. I couldn't have asked for better.

I've given some low-level feedback, but it's up to your discretion to see which points you want to work on. I've been purposefully nitpicky here (since I personally tend to find that more useful), so feel free to disregard anything you feel like isn't a good use of your time.

Again, most of these are minor Rust-related conventions. Let me know if you have any questions!

Thanks again Peter!

chariot-client/src/audio/thread/mod.rs
potor10: Yup! I actually found this as well while doing research on the lib.
The issue with this function, I believe, is that it fades out from the end of the audio source. 

This means that regardless of what is playing, we won't experience a fade out until the audio has reached the end. In the case of music, this could be minutes away!

That's why I made the fade out automate on another thread, so we can immediate request a fade out for a certain duration when we want.

potor10:
# New Updates

> not necessarily a have-to-fix for this PR, but once you hit like three levels of type nesting, you might wanna create a type alias to refer to it. https://doc.rust-lang.org/reference/items/type-aliases.html

Type alias has been added! Thanks for the tip

> I'm not sure what the difference is between unwrap_or_else -> panic and expect but I think expect is more brief.

I think they are pretty similar, but I've changed them over to `expect()` be consistent.

> for all of the following methods that loop over the threads and call a function, I recommend you just do self.threads.iter_mut().for_each and pass a closure instead. http://web.mit.edu/rust-lang_v1.25/arch/amd64_ubuntu1404/share/doc/rust/html/book/first-edition/iterators.html

Yup, I was able to implement the for_each with a closure on all of those methods

Also, I've actually changed the implementation from `Vec` to `HashMap` for both Track & Thread management.

With HashMap we can use the File names as keys for requesting the buffered audio sources. We can also track & maintain threads in a much more ordered fashion. This actually solves the issue of manipulating singular audio threads. We can do all of that now.

> I think this while loop over an increment variable can be simplified to a for loop over self.threads.iter_mut().
Does that throw an error because of concurrent modification? That makes sense.
In which case, look at using the [retain method defined on Vectors](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.retain).

`retain()` has been implemented on clean, and `drain()` has been implemented on `fade_all_threads`

chariot-client/src/audio/mod.rs
```diff
@@ -0,0 +1,246 @@
...
+    thread.play();
+    self.threads.insert(thread_id, thread);
+
+    return Some(thread_id);
+  }
+
+  // Do a fade-out on all currently playing threads in this manager
+  pub fn fade_all_threads(&mut self, duration: Duration) {
+    // Fade Out All Currently Active Threads
+    self.threads.drain().for_each(|(_id, thread)| {
```
KNOXDEV, line 203: Perfect solution! Wasn't even thinking of it.

chariot-client/src/audio/mod.rs
```diff
@@ -0,0 +1,246 @@
...
+  }
+
+  // Spawns a new thread with preapplied volume & pitch for use
+  pub fn spawnThread(&mut self, ctx: &AudioCtx, source: AudioBuffer, opt: SourceOptions) -> AudioThread {
+    let mut thread = AudioThread::new(self.thread_ct, ctx, source, opt);
+    thread.set_volume(self.volume);
+    thread.set_pitch(self.pitch);
+
+    // Increment our track counter
+    self.thread_ct = self.thread_ct + 1;
```
KNOXDEV, line 128: this feels a little...questionable to me. Why not just use `self.threads.len()`?

chariot-client/src/audio/thread/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+  Spatial(SpatialSink),
+  Standard(Sink)
+}
+
+pub struct AudioThread {
+  thread_id: u64,
+  time_start: SystemTime,
+  volume: f32,
+  pitch: f32,
+  source: Buffered<Decoder<BufReader<File>>>,
```
KNOXDEV, line 27: I notice you don't use the type alias you created here, but you can easily share it!

chariot-client/src/audio/thread/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+        volume: 1.0,
+        pitch: 1.0,
+        source: source,
+        sink: sink,
+        src_opt: src_opt
+      }
+    }
+  }
+
+  pub fn getId(&mut self) -> u64 {
```
KNOXDEV, line 81: gonna reiterate what Alistair said, getters and setters are discouraged in Rust.
https://www.reddit.com/r/rust/comments/d7w6n7/is_it_idiomatic_to_write_setters_and_getters/

KNOXDEV: Instead, consider just making `thread_id` public.

chariot-client/src/audio/thread/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+    let tk_src = fade_src.take_duration(self.src_opt.take_duration);
+
+    // Apply Repeat
+    if self.src_opt.repeat {
+      let rpt_src = tk_src.repeat_infinite();
+      match &self.sink {
+        AudioSinkType::Spatial(sink) => sink.append(rpt_src),
+        AudioSinkType::Standard(sink) => sink.append(rpt_src),
+      }
+      //self.sink.append(rpt_src);
```
KNOXDEV, line 118: just remove these comments when you get the chance. Comments are good, commented-out code doesn't need to be there.

chariot-client/src/audio/thread/mod.rs
KNOXDEV: > Yup! I actually found this as well while doing research on the lib. The issue with this function, I believe, is that it fades out from the end of the audio source.
> 
> This means that regardless of what is playing, we won't experience a fade out until the audio has reached the end. In the case of music, this could be minutes away!

I see, but doesn't `take_duration()` accept a duration to end after? The impression I got from that is that you should be able to say "end in two seconds" when do you that.

KNOXDEV (changes requested):
Minor points, Peter. Looking very good!

TheCrossBoy:
Update:
- @potor10 will be working tonight to get more changes in!
- Hopefully we can get it integrated into the client soon as well
- I am able to provide feedback once it's in a satisfactory state!

chariot-client/src/audio/thread/mod.rs
potor10: Hmm, according to rodio docs, `take_duration()` returns a `TakeDuration<>` Source, which means in order for us to use `take_duration()` we cannot give up ownership of the source (which appending to a sink will do). 
https://docs.rs/rodio/latest/rodio/struct.Sink.html#method.append

This means our best attempt at doing a fade_out using this method would be to store another copy of a source, track how long the thread has been alive, and use `skip_duration()` followed by `take_duration()` and the fade_out. 

So we do `stop()` to empty the sink, before doing `append(source.skip_duration().take_duration().set_filter_fadeout())` Generally, this usually results in a super choppy mess, and it doesn't really work at all.

I don't believe there are no inbuilt functions in sink to track the current playtime of a sample, so the best we can do is track the lifetime of the thread, which generally tends to be super inaccurate and results in serious cuts and chops.

chariot-client/src/audio/mod.rs
potor10: I updated it in the most recent commit to be more clear about what it does
`self.thread_ct` has the purpose of tracking the maximum thread id that we currently have out. 
We need to store this value, because there is a chance that using a random `u64` in a hashtable will give us a collision with a currently existing thread. 

This value increases whenever there is no available thread ids for free use, and assigns itself to the newest generated threads.

Previous Implementation: A random unordered Vec<AudioSource> which prevented us from grabbing a specific audio thread from the currently playing group. We could assign IDs to each thread (which I eventually did anyways) but to get a specific thread would be O(n) runtime

Current Implementation: A hashtable storing u64 id key with AudioSource value pairs . A user can request a specific key to obtain the requested pair in O(1) time.

chariot-client/src/audio/thread/mod.rs
KNOXDEV: Hey Peter, the documentation on this is super weak but [by diving into the source I was able to find what Sink does to modify the source after playing has started](https://docs.rs/rodio/latest/src/rodio/sink.rs.html#74).

TLDR, it uses the PeriodicAccess filter with a closure that captures some of the environment to set the source to the configuration settings of the Sink every 5 millisecons. I think this should be the approach you emulate since you know it will work.

chariot-client/src/audio/mod.rs
KNOXDEV: Oh that's cool.
I think Josh / Nick had to solve this same problem in the resource manager. See if this would also solve your problem, since it seems a little less error prone: https://github.com/ucsd-cse125-sp22/cse125-sp22-group1/blob/main/chariot-client/src/resources/mod.rs#L83

KNOXDEV:
One last comment, can you look into adding all these sounds to the [Chaired Drive](https://drive.google.com/drive/u/0/folders/15lMRdrPnF4swE1D9H9jeMP6LyWI5xRbT)? I think we should check the audio into there for consistency rather than committing them to main. 

Additionally, is it possible to use ogg as the audio format? Its pretty popular for games as I understand it.

potor10:
I'll see if I can run the audio files through an online converter or ffmpeg

TheCrossBoy (approved):
as the great alistair says, shock em

PR #43, opened 2022-05-11
State management

trulyronak:
Builds on code from @alistairgray42 to organize state management for the game


chariot-server/src/game/mod.rs
```diff
@@ -0,0 +1,295 @@
...
+                    connection.push_outgoing(ClientBoundPacket::LocationUpdate(locations));
+                }
+            }
+
+            GamePhase::AllPlayersDone(_) => todo!(),
+        }
+    }
+}
+
+pub struct Map {
```
TheCrossBoy, line 297: prob should move map to its own file within the `game` module, just so it's not lost here

chariot-server/src/game/voting.rs
```diff
@@ -0,0 +1,97 @@
...
+                    conn.push_outgoing_message(WSAudienceBoundMessage::Prompt(
+                        state.current_question.clone(),
+                    ))
+                }
+            }
+        }
+    }
+
+    // prompts for a question
+    pub fn start_audience_voting(
```
TheCrossBoy: Copied from earlier:

*TheCrossBoy*
> is it possible to have this be an arbitrary number of options (in a vec) instead of 4? i.e. question: String, options: Vec, poll_time: Duration?
> I think it should be possible to send that to the websocket as well by explicitly stating the number of options, then reading that many, right?
> This is probably not something added in this PR but this is when I'm seeing it most recently, so

*trulyronak*
> I'm going with 4 hard coded options just because I don't want the UI to render more than 4 options → we could do an array of fixed size 4 (which is what it is on typescript actually) but idk how force array size with rust types


*KNOXDEV*
> `options: [String; 4]`

I think removing the strict option number would be preferable, so that we can just do something like `vote("capitalism?", vec!["good", "bad"], 1)` or `vote("pick a number", vec!["1", "2", "3"], 1)`, and we can just sort of not ever have a case where we would try to have more than 4 options

But if this is a limitation with UI on the site, I can help with that as well as there should be an easy (ish) way to solve that with just making stuff flexboxes

chariot-server/src/game/phase.rs
```diff
@@ -0,0 +1,72 @@
...
+#[derive(Debug)]
+pub enum GamePhase {
+    WaitingForPlayerReady(WaitingForPlayerReadyState),
+    CountingDownToGameStart(CountingDownToGameStartState),
+    PlayingGame(PlayingGameState),
+    AllPlayersDone(AllPlayersDoneState),
+}
+
+#[derive(Debug)]
+pub enum VotingState {
```
TheCrossBoy, line 41: I think we may want to alter the voting states (as we may have other cases not covered here) to the following (names TBD):

- `NoVoteActive`: No vote is happening because the we aren't actually playing yet
- `VoteCooldown(uint)`: No vote happening (either game just started, we just finished one, or something else)
- `WaitingForVotes(WaitingForVotesState)` (possibly rename to `VoteActive`, or something similar): Vote has been revealed, we are collecting votes
- `VoteResultActive(uint)`: Whatever the result of the previous vote was is currently happening (different from a cooldown because this is telling us that the previous effect is still happening, and we may want to display that differently)

State progression:
--Server Starts-> `NoVoteActive` --Race Starts-> `Cooldown(seconds)` --`seconds` pass-> `VoteActive(state)` --Vote finishes-> `VoteResultActive(result)` --effect finishes-> Loop @ `Cooldown(seconds)`

While not entirely necessary, it may make it easier to fit this all in in the future.

TheCrossBoy (changes requested):
thank you ronak, very cool

chariot-server/src/game/voting.rs
trulyronak: eh good point we'll just have a bunch in there worst case

chariot-server/src/game/mod.rs
trulyronak: going to let @alistairgray42 do this since I think he's touching this in his own PR

chariot-server/src/game/phase.rs
trulyronak: i like this more i am doing this now

trulyronak: wait question do you mean Instant or Duration for votecooldown

trulyronak: re: NoVoteActive → I don't really see when this is needed so I'm not going to add it right now

TheCrossBoy: @trulyronak for `VoteCooldown`, it would probably make more sense for it to be an `Instant`, so we always know when its supposed to be ending within that state

re `NoVoteActive`, I realize you're right, it's not needed since the vote state is stored within the enum now

TheCrossBoy (approved):
thank you ronak very cool but for the second time

PR #45, opened 2022-05-11
Progress tracking

alistairgray42:
- Keeps track of which players are in which position
- Sends PlacementUpdate packets from server to client whenever players change positions

Will not actually do anything until the map loads in with the checkpoints, since the strategy this uses to keep track of who's in what position depends on what the last checkpoint players have crossed through is. Unfortunately, this means that (because of the way checkpoints work), players will be sorted based on _the farthest they have gone_ - if for some reason a player turns around and goes backwards, this won't update their placement accordingly. The alternative to an approach like this is to use the shape of the actual track and build like a chain of segments along it which we can project locations onto, but that sounds like a lot of work and this sounds like code that should probably work


TheCrossBoy:
> if for some reason a player turns around and goes backwards, this won't update their placement accordingly

I think this actually could work! So the "major checkpoints" will always be the furthest checkpoint you've seen, but the "minor checkpoints" will be fairly common, and will be whatever checkpoint the one you were most recently in was. It isn't used for tracking lap progress, but just as a method of knowing where you are.

Downside is twofold:
- We have to add the minor checkpoints to the map (shouldn't? be that hard? I think it's as simple as making a bunch of boxes in blender which *shouldn't* be that hard hopefully)
- Two players in the same minor checkpoint are considered in an equal place, but this should also be fine (would look like a slight delay in placement updating, which honestly shouldn't be *that* noticeable anyways)

alistairgray42:
>     * Two players in the same minor checkpoint are considered in an equal place, but this should also be fine (would look like a slight delay in placement updating, which honestly shouldn't be _that_ noticeable anyways)

aha! there's one more level of granularity beyond the minor checkpoints (`position_within_checkpoint`) to try to deal with this, which is linearly interpolated between checkpoints based on the projection of each player's position onto the line between the last minor checkpoint they've hit and the next one in line.

chariot-server/src/checkpoints.rs
```diff
@@ -7,7 +7,7 @@ type BoundingBoxDimensions = [[f64; 2]; 3];
 #[derive(Clone, Copy)]
 pub struct MinorCheckpoint {
     pub id: MinorCheckpointID,
-    pos: DVec3,
+    pub pos: DVec3,
     size: DVec3,
```
trulyronak, line 11: any reason not to make size `pub` as well at this rate?

chariot-server/src/game/mod.rs
```diff
@@ -291,16 +294,31 @@ impl GameServer {
...
             GamePhase::AllPlayersDone(_) => todo!(),
         }
+
+        if let Some(map) = &self.map {
+            if let PlayingGame(state) = &self.game_state.phase {
+                let old_placement_array = state.player_placement;
+                let new_placement_array =
+                    get_player_placement_array(&self.game_state.players, &map.checkpoints);
+
+                for index in 0..=3 {
```
trulyronak, line 313: is this hardcoded for 4 players? there's some ambiguity as to what the 0-3 means (can you rename `index` to be  more descriptive)

chariot-server/src/progress.rs
```diff
@@ -0,0 +1,63 @@
...
+    fn get_progress_score(&self, minor_checkpoints: &Vec<MinorCheckpoint>) -> (u8, u8, u8, f64) {
+        // There's a hierarchy of four pieces of progress information we have:
+        // from least to most granular lap number, zone number (zones are
+        // between major checkpoints), minor checkpoint number, and linear
+        // interpolation between minor checkpoint
+        let lap_number = self.lap_info.lap;
+        let zone_number = self.lap_info.zone;
+        let checkpoint_number = self.lap_info.last_checkpoint;
+
+        let current_checkpoint = minor_checkpoints.get(checkpoint_number as usize).unwrap();
```
trulyronak, line 14: can you change these `unwraps` to be `expects`, since (a) this will fail if it breaks anyways and (b) it shouldn't fail to unwrap here

chariot-server/src/game/phase.rs
```diff
@@ -65,6 +65,7 @@ pub struct CountingDownToGameStartState {
 #[derive(Debug)]
 pub struct PlayingGameState {
     pub voting_game_state: VotingState,
+    pub player_placement: [u8; 4],
```
trulyronak, line 68: any reason you're not using a `u32` or `i32`? seems like you have to do a bunch of casting later anyways

trulyronak (approved):

trulyronak (commented):
meant to do a comment

chariot-server/src/game/phase.rs
alistairgray42: mostly just range/aesthetics reasons, but actually some of those casts turned out to be unnecessary so yay

PR #46, opened 2022-05-11
Audience Interaction - it's all coming together

trulyronak:
With this PR, you can now play a game, and join the website and be prompted to decide based on a question (sent via `questions.yaml`)

Provides support for 
- easy question types
- different types of actions (enum)
- doing something for this action type

todo
- remove printlns on client so that we can see what the ui would look like (in a janky terminal kinda way)


trulyronak:
https://627c445fc6c4b31e2927f7b1--chairiot.netlify.app//

chariot-core/src/questions.rs
```diff
@@ -0,0 +1,54 @@
+use lazy_static::lazy_static;
+use rand::seq::SliceRandom;
+use rand::thread_rng;
+use serde::{Deserialize, Serialize};
+use std::env;
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct Questions {
```
alistairgray42, line 8: Is there a reason this struct needs to exist - like, does this add any additional functionality on top of using `Vec<QuestionData>` instead? or could there be any added functionality later? (e.g. moving the index from `GameServer` to here) I think that if neither of those things is true, it'd be simpler not to have this struct at all. alternatively, if you want to keep this, I don't think it needs to derive `Serialize`, `Deserialize`, or `Clone`

chariot-core/src/questions.rs
```diff
@@ -0,0 +1,54 @@
...
+    NoRight,
+}
+
+impl Questions {
+    fn load_questions() -> Questions {
+        // get questions file
+        let questions_yaml_path = format!(
+            "{}/chariot-core/questions.yaml",
+            env::current_dir()
+                .unwrap()
```
alistairgray42, line 35: perhaps a better option for this whole function is returning a `Result` and using the `?` operator to bubble up errors from I/O stuff (since conceptually speaking, this function shouldn't necessarily always succeed) - that'd make this function just less verbose

chariot-server/src/game/mod.rs
```diff
@@ -28,6 +30,7 @@ pub struct GameServer {
     ws_connections: HashMap<Uuid, WebSocketConnection>,
     game_state: ServerGameState,
     map: Option<Map>,
+    question_idx: usize, // to keep track of which question we have asked
```
alistairgray42, line 33: I don't like having variables like this one in `GameServer`, and would prefer to move it downwards, into somewhere like `VotingState`

alistairgray42 (commented):
Not reviewing the web stuff

Looks pretty good to me! Just on a general level I'm concerned that a lot of business logic is accumulating in `game/mod.rs`, but given the fact that that aggregates a lot of our state and such, it seems pretty nontrivial to disentangle that. however this code works and that's pretty cool :rocket:

trulyronak:
Fair comment, I'm thinking I might do some refactoring of that as well, but thinking the approach to do so would become more apparent when we write more game logic

KNOXDEV:
image
wow Ronak really???

KNOXDEV:
Also I can't really review this while there are still conflicts, and since you said you would handle them, I'll wait for that. Again, if you want me to take care of it just say the word.

chariot-core/src/questions.rs
trulyronak: it needs serialize/deserialize/clone all just for serde_json parse to work

chariot-server/src/game/mod.rs
trulyronak: moved!

chariot-core/src/questions.rs
trulyronak: hopped on call with @alistairgray42 , we couldn't figure out a clean alternative, and so I just left it as is pretty much

chariot-client/src/graphics.rs
```diff
@@ -171,6 +171,7 @@ impl GraphicsManager {
             self.add_player(player_num, false);
         }
         let player_entity = self.player_entities[player_num as usize].unwrap();
+
```
TheCrossBoy, line 174: Truly a riveting change.

chariot-core/src/questions.rs
```diff
@@ -0,0 +1,54 @@
+use lazy_static::lazy_static;
+use rand::seq::SliceRandom;
+use rand::thread_rng;
+use serde::{Deserialize, Serialize};
+use std::env;
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct Questions {
+    pub questions: Vec<QuestionData>,
```
TheCrossBoy, line 9: I think if `QuestionData` implements those traits, you should be able to just use a vector without needing a struct to contain it, which would be cleaner imo

TheCrossBoy (commented):
Nice!

chariot-core/src/questions.rs
trulyronak: fixed it, had to modify the yaml file

alistairgray42 (approved):

PR #48, opened 2022-05-13
Lots of render changes

TheCrossBoy:
Lots of stuff:

- Moved to a graph traversal for glb loading
  - While probably not fully necessary, it makes determining the transform for something much easier
  - We don't have to stick with this is it seems unnecessary, but it hopefully solves a few issues assuming the kinks are resolved
- No longer limited to 1 texture / mesh (I think?)
- Removed a lot of stuff that was (might have been?) redundant
  - Creating extra player chair
  - Some returns that were unnecessary
  - Extra looping in loading
- Removed a lot of the manual transforms we were doing to fix the fact we weren't getting transform data from models
  - Chairs, the world, etc
  - Also, rather than the client shifting the world down, the server spawns players higher (to match the real position)
- QOL stuff
  - Moved resources to chariot-core, so server can access them too
  - You can press escape to exit the game

Current problems:
- ~~Shadows might be a bit weird, sometimes they look off but might just be how they are~~
- ~~Geometry nodes don't work unless applied~~
  - ~~Maybe this is how it's supposed to work?~~
  - ~~It seems dumb that that's how it'd have to work but I'm not certain what I'm doing wrong here~~
  - ~~I believe the fully baked file looks the same, but the unbaked ones do have this issue~~
- Probably other things I'm forgetting

### DIFF
--- a/chariot-client/src/graphics.rs
+++ b/chariot-client/src/graphics.rs
@@ -27,29 +27,17 @@ pub fn register_passes(renderer: &mut Renderer) {
     );
 }
 
-fn setup_world(resources: &mut ResourceManager, renderer: &mut Renderer) -> (World, Entity) {
+fn setup_world(resources: &mut ResourceManager, renderer: &mut Renderer) -> World {
     let mut world = World::new();
     world.register::<Camera>();
     world.register::<Vec<StaticMeshDrawable>>();
     world.register::<Bounds>();
     world.register::<Light>();
     let world_root = world.root();
-    let chair = {
-        let chair_import = resources
-            .import_gltf(renderer, "models/defaultchair.glb")
-            .expect("Failed to import chair");
 
-        world
-            .builder()
-            .attach(world_root)
-            .with(Transform::default())
-            .with(chair_import.drawables)
-            .with(chair_import.bounds)
-            .build()
-    };
     {
         let track_import = resources
-            .import_gltf(renderer, "models/baked-triggered.glb")
+            .import_gltf(renderer, "models/track.glb")
             .expect("Unable to load racetrack");
 
         let track = world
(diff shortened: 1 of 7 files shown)


KNOXDEV:
Heads up that you might have a little bit of work to do after the camera PR. I'll add some comments identifying the points of contention, and I'll also check out the refactors

chariot-client/src/main.rs
```diff
@@ -57,6 +57,11 @@ fn main() {
                 event: WindowEvent::KeyboardInput { input, .. },
                 ..
             } => {
+                // TODO: Make this be a more real thing, rather than hard coding it here
```
KNOXDEV, line 60: this seems like enough of a "real thing" to me.

chariot-client/src/graphics.rs
```diff
@@ -212,11 +192,13 @@ impl GraphicsManager {
...
             player_num, location.unit_upward_direction
         );
+        */
         let player_transform = self
             .world
             .get_mut::<Transform>(player_entity)
             .expect("Trying to update player location when transform does not exist");
-        *player_transform = GraphicsManager::EntityLocation_to_Transform(&location);
+        *player_transform =
+            GraphicsManager::EntityLocation_to_Transform(&location, player_transform.scale);
```
KNOXDEV, line 201: Note that in my PR I moved this function into `Transform::from_entity_location`.

KNOXDEV (commented):
Without having run it, it looks like a bunch of good improvements to me.

Deferring to Nick for the approval.

nickwn (dismissed):
Looks nice! In the future it would be cool to have import return a World (and add a merge function for worlds) so that way it has hierarchy info and we’re not duplicating instanced models, but that’s a job for another day.

chariot-client/src/main.rs
TheCrossBoy: I wasn't the biggest fan of it just being a one off check outside of the rest of our input handling, but it's probably fine since otherwise we'd need to pass around the windows variables or get data back up here, so agreed

nickwn (approved):

PR #49, opened 2022-05-13
aaron/camera facelift

KNOXDEV:
Mildly smoother camera, mainly. Try it out.


alistairgray42 (approved):
feels better to me! camera is jumpy when braking and near/at a standstill, but i think that's a physics issue

PR #50, opened 2022-05-13
Physics fixes

alistairgray42:
Fixes three physics issues (one commit per issue) for better QoL:

- Colliding chairs kinda jittered together / didn't leave room for Jesus before separating when colliding: this turns out to be because the collision force was facing the wrong direction, so weird things happened
- Occasionally, collisions (especially when 3 chairs were involved) added a vertical component, so chairs went above or below the plane of the track
- The camera was very flickery / changed directions unexpectedly when velocity was very low; the solution here consisted of zeroing out velocity when under a certain threshold so that the camera direction wouldn't change in response to very small velocities, not changing the camera angle at all when the client is given a velocity of zero (necessary since otherwise the camera resets its angle to not depend on velocity), and adding an edge case to not break everything when braking if velocity is zero

All of this was tested live and looks good - in particular, collisions (with Aidan from Group 4, shout out) seem not unreasonable, but acceleration is fast enough that it's not very impactful to actually collide with another player.

### DIFF
--- a/chariot-client/src/graphics.rs
+++ b/chariot-client/src/graphics.rs
@@ -198,7 +198,7 @@ impl GraphicsManager {
         *player_transform = Transform::from_entity_location(&location);
 
         // if this player is the main player, update the camera too (based on velocity)
-        if player_entity == self.camera_entity {
+        if player_entity == self.camera_entity && *velocity != DVec3::ZERO {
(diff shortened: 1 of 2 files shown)


KNOXDEV (approved):
I dig it, but when adding code that handles edge cases, just leave a comment explaining why its there so we don't forget later.

KNOXDEV (approved):

PR #51, opened 2022-05-14
added simple ui drawable

nickwn:
+ some fixes to framebuffers without depth
Usage to create is: 
```rust
let text_handle = resources.import_texture(&renderer, "text.png");
let text_tex = resources.textures.get(&text_handle).unwrap();
let test_ui = UIDrawable {
    // layers specifies a list of draws that will be done in order to get proper foreground/backround
    layers: vec![technique::UILayerTechnique::new(
        &renderer,
        glam::vec2(0.0, 0.0), // pos on screen (0,0 is upper left)
        glam::vec2(0.2, 0.2), // size on screen (normalized where 1, 1 is size of entire screen)
        glam::vec2(0.0, 0.0), // texture coord pos (normalized)
        glam::vec2(1.0, 1.0), // texture coord size, (normalized, where 1,1 is entire texture)
        &text_tex,
    )],
};
```
Then the draw graph is merged after postprocessing in `render()`.
A little awkward with all the options but it should be able to handle a good range of things.

### DIFF
--- a/chariot-client/src/drawable/technique.rs
+++ b/chariot-client/src/drawable/technique.rs
@@ -1,4 +1,5 @@
 use crate::renderer::*;
+use crate::resources::material::MaterialBuilder;
 use crate::resources::*;
 use wgpu::util::DeviceExt;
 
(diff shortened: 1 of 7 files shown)


chariot-client/src/drawable/technique.rs
```diff
@@ -269,3 +270,93 @@ impl Technique for FSQTechnique {
...
+
+pub struct UILayerTechnique {
+    vertex_buffer: wgpu::Buffer,
+    texcoord_buffer: wgpu::Buffer,
+    index_buffer: wgpu::Buffer,
+    material: material::Material,
+}
+
+impl UILayerTechnique {
+    pub fn new(
```
TheCrossBoy, line 282: copying from discord:

a solution to all the options (assuming it was something we wanted to address) might be to have ``UILayerTechnique::new` have a signature of `Renderer, Texture -> UILayerTechnique`, and fill in the vecs with default values, then just add `.size(vec2)`, `.pos(vec2)`, `.textcordpos(vec2)`, etc (might even be easier to do `(float, float)` instead which all just return self with the modification, that way you can specify just what you want and the default stuff that usually isn't used as much isn't needed

TheCrossBoy (approved):
very neat, thing I mentioned is really optional/QOL for the future so not totally needed. very good stuff here!

chariot-client/src/drawable/technique.rs
nickwn: I could but in short I'm too lazy, it would require a full on builder struct to keep the reference to renderer so you're not passing it in every time you're calling one of the helpers (.pos(), .size()) to upload a new buffer with some changes.

PR #52, opened 2022-05-15
Create dynamic minimap

alistairgray42:
not quite ready yet! but almost!

What this can currently do:

- Put dots on a transparent track map showing where all players are, and draw this to the screen
- run it! you will see dots in the right places
- we're going to want some kind of nice UI element to put this on top of / will want to mess with colors and transparency and such so that it's easily readable

Still needs to be done:
- calibration of location conversion - when e.g. under Geisel the location on the map looks wrong
- make this look nicer? like it currently looks pretty squashed and the dots are small and could do with a border, things like that


chariot-client/src/graphics.rs
```diff
@@ -225,6 +226,51 @@ impl GraphicsManager {
...
+                    .get::<Transform>(player_num.unwrap())
+                    .unwrap()
+                    .translation;
+                (location.x, location.z)
+            })
+            .collect();
+
+        let minimap = create_minimap_image(player_locations, &mut self.resources);
+
+        let texture = self.renderer.create_texture2D_init(
```
alistairgray42, line 250: I just kinda copied this stuff from the text example; Nick, is it as bad of an idea as it sounds to create a new texture and UI layer each renderer tick?

nickwn: If it's real time then sure but I would try not to. Sorry for mentioning this now, but I think it actually might be easier to have the track and all players as their own UILayerTechnique, then update their vertex buffers each frame to move them around and let the gpu composite everything when rendering. The vertex buffer can be updated using something like `renderer.write_buffer(ui_layer.vertex_buffer, data)`. The UILayerTechniques for the individual players could just use a 1x1 texture for now.

TheCrossBoy:
Updates:
- Currently waiting on the changes mentioned by nic and other visual tweaks
- Lower priority than terrain collisions
- Will be picked back up after the above!

chariot-client/src/graphics.rs
```diff
@@ -284,8 +330,51 @@ impl GraphicsManager {
...
+                    .get::<Transform>(player_num.unwrap())
+                    .unwrap()
+                    .translation;
+                (location.x, location.z)
+            })
+            .map(get_minimap_player_location);
+
+        for (player_index, location) in player_locations.enumerate() {
+            let player_layer = self.minimap_ui.layers.get_mut(player_index + 1).unwrap();
+            let new_vertex_buffer = UILayerTechnique::create_vertex_buffer(
```
alistairgray42, line 368: @nickwn this feels somewhat less silly (but silly kinda silly tbh); thoughts?

alistairgray42:
image

chariot-client/src/graphics.rs
```diff
@@ -105,6 +106,7 @@ pub struct GraphicsManager {
     pub renderer: Renderer,
     pub resources: ResourceManager,
 
+    minimap_ui: UIDrawable,
```
TheCrossBoy, line 109: We're starting to have a lot of UIDrawable things start to interact with one another. Perhaps we should have some kind of `Vec<UIDrawable>` to keep track of this, that way we don't have to edit the struct every time we want to add something?

This may be added in the UI PR, but it seemed pertinent to mention it here at least!

chariot-client/src/graphics.rs
```diff
@@ -284,8 +330,51 @@ impl GraphicsManager {
...
+            // these values are guesses btw
+            const MIN_TRACK_X: f32 = -113.0;
+            const MAX_TRACK_X: f32 = 37.0;
+            const MIN_TRACK_Z: f32 = -39.0;
+            const MAX_TRACK_Z: f32 = 111.0;
+
+            (
+                (MAX_TRACK_Z - location.1) / (MAX_TRACK_Z - MIN_TRACK_Z),
+                (location.0 - MIN_TRACK_X) / (MAX_TRACK_X - MIN_TRACK_X),
+            )
```
TheCrossBoy, line 350: We should be able to (hypothetically) get more exact values from this, but it's probably close enough for now

chariot-client/src/graphics.rs
```diff
@@ -284,8 +330,51 @@ impl GraphicsManager {
...
+            .map(get_minimap_player_location);
+
+        for (player_index, location) in player_locations.enumerate() {
+            let player_layer = self.minimap_ui.layers.get_mut(player_index + 1).unwrap();
+            let new_vertex_buffer = UILayerTechnique::create_vertex_buffer(
+                &self.renderer,
+                Vec2::new(0.2 * location.0, 0.2 * location.1),
+                Vec2::new(0.02, 0.02),
+            );
+            player_layer.vertex_buffer = new_vertex_buffer;
```
TheCrossBoy, line 373: (I can't add a add this comment to my review without having it not respond to yours, so pretend I'm responding to your comment)
You shouldn't have to create a whole new buffer, you should be able to do something like `renderer.write_buffer(player_layer.vertex_buffer, verts_data [from UILayerTechnique::create_vertex_buffer])`, and it should update the buffer without having to create a whole new one

TheCrossBoy (changes requested):
Good stuff! I will note it is slightly off, but I think it's close enough that for now it should be okay: 

Just see if you can get the `write_buffer` to work and this will be all thumbs up from me!

nickwn (approved):
Looks good, but it would be nice to have a talk with everyone on how the UI should be organized, since I don't think its healthy to keep adding to graphics.rs

chariot-client/src/graphics.rs
alistairgray42: i made it somewhat close enough-er but it's very annoying to mess with these values so i'm done with that for now

trulyronak (approved):
bussing

TheCrossBoy (approved):
Very cool :)

PR #53, opened 2022-05-17
fixed bug where voting wasn't looping

trulyronak:
before
- after the first vote, we didn't do anything

now
- after the first vote, we go to `VoteCooldown`, wait 10 seconds, then start up voting again with a new question


alistairgray42 (approved):
🚀

PR #55, opened 2022-05-17
font rendering probably

KNOXDEV:
Okay here's the current plan:

- [x] use fontkit to import fonts from the system and rasterize glyphs
- [x] create a glyph cache that will manage wgpu textures containing rasterized glyphs and allocate new textures when more are required
- [x] create a string cache that will store vectors of layout information (probably as UILayerTechniques)
- [x] create an api by which to return a UIDrawable when given a string to render, leveraging the string cache


chariot-client/src/renderer/mod.rs
```diff
@@ -30,7 +30,7 @@ use render_job::*;
  *				bind uniform set 1.1	| bind_group[0]		| mvp buffer			|
  *				bind uniform set 1.2	| bind_group[1]		| material data buffer	|
  *					draw()
- * These next two are skipped becase they are the same as the above
+ * These next two are skipped because they are the same as the above
```
TheCrossBoy, line 33: HUGE

KNOXDEV:
Kind of disorganized rn but I just want somebody else to look at it before i explode

chariot-client/src/resources/static_mesh.rs
```diff
@@ -48,7 +48,7 @@ impl<'a> MeshBuilder<'a> {
         }
     }
 
-    pub fn vertex_buffer<'b, T: bytemuck::Pod>(&'b mut self, data: &[T]) -> &'b mut Self {
```
KNOXDEV, line 51: Whoops my IDE removed all these unnecessary lifetimes please disregard

chariot-client/src/renderer/render_job.rs
```diff
@@ -192,9 +192,10 @@ impl<'a> RenderGraphBuilder<'a> {
 
         if deps.is_empty() {
             self.render_graph.roots.push(res_id);
-            self.render_graph.nodes.insert(res_id, vec![]);
         }
 
+        self.render_graph.nodes.insert(res_id, vec![]);
```
KNOXDEV, line 197: @nickwn or maybe @TheCrossBoy idk but I think this was a bug in the render graph builder and I THINK I fixed it

KNOXDEV:
Okay beyond that, it is supposed to be cross platform but if it dies on your mac lmk

chariot-client/src/renderer/render_job.rs
nickwn: Yep, I actually fixed this in my indirect branch as well, sorry about that

chariot-client/src/drawable/string.rs
```diff
@@ -0,0 +1,68 @@
...
+use crate::resources::ResourceManager;
+use glam::Vec2;
+use std::ops::{Add, Div};
+
+pub struct StringDrawable {
+    ui_drawable: UIDrawable,
+    glyph_cache: GlyphCache,
+}
+
+impl StringDrawable {
```
nickwn, line 14: 1. Does each string drawable need its own copy of a glyph cache? It might be better to have the glyph caches owned by the resource manager and access it through handles.
2. If the glyph cache can be removed as a field, this could be turned into a StringDrawableBuilder or something where it just returns a UIDrawable.

chariot-client/src/renderer/render_job.rs
TheCrossBoy: im pretty sure this is a much better way of fixing the bug that i also fixed so assuming there's not a side effect of this +1 to me

chariot-client/src/font/glyph_cache.rs
```diff
@@ -0,0 +1,250 @@
...
+        character: char,
+        renderer: &Renderer,
+        resource_manager: &mut ResourceManager,
+    ) -> &Glyph {
+        // fetch the glyph_id for this character
+        // TODO: rather than CRASH, we should render the "unrecognized character" glyph
+        let glyph_id = self
+            .font
+            .glyph_for_char(character)
+            .expect("glyph for character unavailable");
```
TheCrossBoy, line 120: i vote for just never try to do it and therefore it will never be a bug to fix :)

chariot-client/src/graphics.rs
```diff
@@ -146,33 +146,28 @@ impl GraphicsManager {
...
+            &renderer,
+            &mut resources,
+        );
 
         let postprocess = technique::FSQTechnique::new(&renderer, &resources, "postprocess");
 
         let world = setup_void(&mut resources, &mut renderer);
 
         Self {
+            test_string,
```
trulyronak, line 162: question - would we have an array of all the strings we are rendering on the screen stored here or something? 

also, how does removing a string to render work? (would we just simply not call 

```
let text_graph = self.test_string.render_graph(&self.resources);
        render_job.merge_graph_after("postprocess", text_graph);
```
then? (thats why i mentioned it in discord)

chariot-client/src/graphics.rs
```diff
@@ -407,8 +402,8 @@ impl GraphicsManager {
         let postprocess_graph = self.postprocess.render_item(&self.resources).to_graph();
         render_job.merge_graph_after("forward", postprocess_graph);
 
-        let ui_graph = self.test_ui.render_graph(&self.resources);
-        render_job.merge_graph_after("postprocess", ui_graph);
+        let text_graph = self.test_string.render_graph(&self.resources);
+        render_job.merge_graph_after("postprocess", text_graph);
```
trulyronak, line 406: if these two lines don't happen, does this mean the text doesn't render? is this basically just the communication to do the specific rendering stuff itself?

trulyronak (commented):
looks pog, just a question re removing text + storing text

TheCrossBoy (approved):
very cool thank you :)

your sanity was not lost in vain my friend

chariot-client/src/graphics.rs
TheCrossBoy: yup, the render graph is how the renderer figures out what to render. by putting something into the render graph, we tell it "hey we need you to figure out how to draw this", in this case, we do it under "postprocess", which means after the postprocessing is done, it will figure out how the text_graph adds into it.

chariot-client/src/graphics.rs
TheCrossBoy: The render graph stuff is called on every frame, which means if we modify the contents of the tests_string (or other things it depends on), it should be updated per frame

chariot-client/src/drawable/string.rs
KNOXDEV: Each unique font+point size pair needs its own GlyphCache because of how it handles updating the texture.
Knowing that, what would you suggest?

nickwn: Looks fine then. Maybe the resource manager could have a hash map of (font, size) to glyph cache but at that point if each string is likely to be different (and we don't have too many strings) then it wouldn't be too bad to just store it on the drawable. Either way is good with me

KNOXDEV: I took a look at this solution because I think its a good one, but [as it turns out](https://stackoverflow.com/questions/45786717/how-to-implement-hashmap-with-two-keys), its not so simple to build such a hashmap. we also run into fun problems where ResourceManager owns all the GlyphCaches but GlyphCache needs to borrow ResourceManager...

I'll leave it the way it is for now since the API isn't likely to change and we can hurry and merge this in.

PR #56, opened 2022-05-17
Having a real game to play!

TheCrossBoy:
Going to be working on the major hurdles preventing us from having a real game going. Mostly server side, client will be informed but visuals for it will likely not be on this PR.

- [x] Separate states for choosing playing parameters vs loading into the game
- [x] Let the client actually pick a chair
- [x] See if connections to the server can be handled non-blocking-ly, so we can choose chairs and maps before everyone has loaded
- [x] Restructure connecting logic so less stuff is required beforehand
- [x] Don't load the map on the client until the server tells you to
- [x] Load specific chairs for players, rather than the default one
- [x] Spawn players into the map ready to race
- [x] After everyone has completed the race, be in a "post game" view
- [x] Return to the parameter selection screen after

Will be trying to work on this as much as possible over the next few days, because it's sort of the basis of... well everything.


chariot-server/src/game/phase.rs
```diff
@@ -11,7 +11,7 @@ use super::voting::{AnswerID, QuestionID};
 /*
  * Phases of the game are as follows:
 *
- * 1. GamePhase::WaitingForPlayerReady
+ * 1. GamePhase::ChoosingSettingsAndConnecting
```
trulyronak, line 14: shouldn't there be an earlier stage, waiting for players to connect? because until the players connect, we can't even get to this stage

TheCrossBoy:
Alright as of now everything seems to be working from (Start server/client) -> (post-game), the transition from (post-game) -> (pre-game) is the only thing not working.
Looks like there will need to be more work done to ensure the previous stuff is "un-rendered", and the server is properly able to accept new stuff from the client. 

also should make sure the server unloads the map as well, just thought of that too

chariot-client/src/application.rs
```diff
@@ -123,6 +176,21 @@ impl Application {
         if key == VirtualKeyCode::R {
             println!("Reloading shaders");
             register_passes(&mut self.graphics.renderer);
+        } else if key == VirtualKeyCode::Return {
+            println!("Picking chair");
+            self.game.pick_chair("standard".to_string());
```
trulyronak, line 181: should make all the chairs an enum type

chariot-client/src/game.rs
```diff
@@ -15,23 +15,40 @@ impl GameClient {
...
 
-    pub fn send_ready_packet(&mut self, chair_name: String) {
-        self.connection
-            .push_outgoing(ServerBoundPacket::ChairSelectAndReady(chair_name));
+    fn send_packet(&mut self, packet: ServerBoundPacket) {
+        self.connection.push_outgoing(packet);
         self.connection.sync_outgoing();
     }
 
+    pub fn pick_chair(&mut self, chair: String) {
```
trulyronak, line 27: would be best to make chair a shared enum

chariot-client/src/game.rs
```diff
@@ -15,23 +15,40 @@ impl GameClient {
...
+        self.send_packet(ServerBoundPacket::ChairSelect(chair));
+    }
+
+    pub fn pick_map(&mut self, map: String) {
```
trulyronak, line 31: similar, maps should be an enum

chariot-client/src/graphics.rs
```diff
@@ -132,25 +160,61 @@ impl GraphicsManager {
...
+            postprocess,
+            player_num: 4,
             camera_entity: NULL_ENTITY,
         }
     }
 
-    pub fn add_player(&mut self, player_num: PlayerID, is_self: bool) {
+    pub fn load_menu(&mut self) {
+        println!("Loading main menu!");
+        // todo!();
```
trulyronak, line 182: nit: shouldn't keep any commented code

chariot-core/src/settings.rs
```diff
@@ -46,12 +48,13 @@ impl Settings {
...
-            .set_default("player_amount", 1)?
-            // for debugging purposes: don't require all players to join to get to see the graphical view
-            .set_default("bypass_multiplayer_requirement", false)?
+            .set_default("player_amount", 4)?
             // files_and_resources (client)
             .set_default("resource_folder", "../chariot-core/resources")?
-            .set_default("map_name", "track")?
+            // Defaults
+            .set_default("default_chair", "standard")?
+            .set_default("default_map_vote", "track")?
```
trulyronak, line 56: don't think this needs to be customized in settings, it's already changable by the players + by the default implementation

trulyronak (changes requested):
pretty much good to go, added some comments (that i will resolve myself) to improve maintainabiliy + readability

trulyronak (approved):
a little biased, but this lgtm!

PR #57, opened 2022-05-18
WS Secure w/ SSL

trulyronak:
I really hate SSL, but this should work, barring invalid certs


TheCrossBoy:
Updates
- This is a production issue, so much lower priority atm!
- Ronak needs to meet with Aaron to get the last bit solved
- Ronak is _also_ avoiding people because plague is very pog

PR #59, opened 2022-05-19
Add collider loading/support for maps on server

TheCrossBoy:
Should load and probably collide. Requires the entity to have the object data property `purpose` set to `collider`. 

Currently `track.glb` doesn't have this set at all, so you'll need to fire up the ol' blender to get somethin in there. On my laptop so I couldn't test that part.

Feel free to take over if you'd like/don't have anything, otherwise you can review and I'll get to it when I can (will be at least a few hours till I'm home though)


alistairgray42:
Ready for review. Chairs can reflect off objects now! that's cool!

Along with this PR, i pushed blend and gltb files with various buildings in the scene set to be collideable (custom data property purpose set to collider). Unfortunately, because of AABB there's a few issues with this that arise from concavity / three-dimensionality:
- Geisel tunnel can't be traversed
- Sun God takes up way more space than it should based on appearance
- Same with the concave bit in the smaller side of PC
- Same with the concave parts of Jacobs

I think the way to fix this is by creating dummy collider objects in Blender that actually have the dimensions we want to be collideable, and making the actual objects non-collideable. I'm not at all comfortable with Blender, so if someone else could handle that I'd be appreciative. The Geisel and PC ones are most important.

(copied from Discord)

alistairgray42 (commented):
passing off to josh for actual review

TheCrossBoy (commented):
Code wise looks good! I can't try it in game now but I'll trust your judgement.

alistairgray42 (approved):

PR #60, opened 2022-05-20
Adding in chair choices

TheCrossBoy:
Going to go based on the chairs we have with models rn

- [x] Swivel chair
  - [x] Balanced stats, get this done _first_
  - [x] Wheels don't turn with the camera
- [x] Arm Chair
  - [x] Turns _very_ slowly, higher top speed, lower acceleration
- [x] Ergonomic
  - [x] High handling, low acceleration
- [x] Beanbag
  - [x] HIGH top speed. VERY LOW drag.

### DIFF
--- a/chariot-client/src/graphics.rs
+++ b/chariot-client/src/graphics.rs
@@ -213,7 +213,7 @@ impl GraphicsManager {
             .resources
             .import_gltf(
                 &mut self.renderer,
-                format!("models/{}.glb", choices.chair).to_string(),
+                format!("models/{}.glb", choices.chair.file()).to_string(),
             )
             .expect("Failed to import chair");
 
--- a/chariot-client/src/graphics.rs
+++ b/chariot-client/src/graphics.rs
@@ -274,8 +274,7 @@ impl GraphicsManager {
                     DVec3::new(velocity.x, 0.0, velocity.z).angle_between(DVec3::X);
 
                 // there's actually some magic trig cancellations happening here that simplify this calculation
-                let mut orbit_yaw = velocity.z.signum() * velocity_angle
-                    - location.unit_steer_direction.z.signum() * rotation_angle;
+                let mut orbit_yaw = PI + location.unit_steer_direction.z.signum() * rotation_angle;
 
                 // if the yaw change would be bigger than PI, wrap back around
                 let yaw_difference = orbit_yaw - camera.orbit_angle.x as f64;
@@ -284,8 +283,7 @@ impl GraphicsManager {
                 }
 
                 // set the new orbit angle complete with magic pitch for now
-                camera.orbit_angle =
-                    Vec2::new(orbit_yaw as f32, -0.3).lerp(camera.orbit_angle, 0.5);
+                camera.orbit_angle = Vec2::new(0.0, -0.3).lerp(camera.orbit_angle, 0.5);
             }
         }
     }
@@ -376,17 +374,28 @@ impl GraphicsManager {
             .map(|l| l.calc_view_proj(&view_bounds))
             .collect();
 
+        let default_translation = Transform::default();
         let mut render_job = render_job::RenderJob::default();
         self.world.dfs_acc(self.world.root(), root_xform, |e, acc| {
-            let cur_model = self
+            let cur_transform = self
                 .world
                 .get::<Transform>(e)
-                .unwrap_or(&Transform::default())
-                .to_mat4();
+                .unwrap_or(&default_translation);
+            let cur_model = cur_transform.to_mat4();
             let acc_model = *acc * cur_model;
 
             if let Some(drawables) = self.world.get::<Vec<StaticMeshDrawable>>(e) {
                 for drawable in drawables.iter() {
+                    let mut acc_model = acc_model;
+                    if drawable.modifiers.absolute_angle {
+                        acc_model = *acc
+                            * Transform {
+                                translation: cur_transform.translation,
+                                rotation: glam::Quat::IDENTITY,
+                                scale: cur_transform.scale,
+                            }
+                            .to_mat4();
+                    }
                     drawable.update_xforms(&self.renderer, proj, view, acc_model);
                     drawable.update_lights(&self.renderer, acc_model, &lights);
                     let render_graph = drawable.render_graph(&self.resources);
--- a/chariot-client/src/graphics.rs
+++ b/chariot-client/src/graphics.rs
@@ -275,6 +275,7 @@ P tells the server to start the next round",
             Transform::from_entity_location(&location, player_transform.scale);
         *player_transform = new_player_transform;
 
+        // If we are moving, we might need to rotate the model
         if velocity.length() > 0.0 {
             if let Some(Chair::Beanbag) = self.player_choices[player_num].as_ref().map(|c| c.chair)
             {
@@ -284,10 +285,21 @@ P tells the server to start the next round",
                     .unwrap()
                     .iter_mut()
                 {
+                    // We need an axis and angle of orientation
+                    // Axis is what we spin on
+                    // Angle is how far we spin
+                    // If we are moving towards velocity, we want to spin "downward" towards it
+                    // Thus, the axis is the "right-left" of velocity
+                    // This is what we get here
                     let axis = velocity.as_vec3().cross(glam::Vec3::Y).normalize();
                     drawable.modifiers.rotation = Some(
                         glam::Quat::from_axis_angle(
                             axis,
+                            // For the angle, we want to move the velocity's length
+                            // But we could go either + or - that amount
+                            // We want to go "towards the ground"
+                            // So we figure out if the angle between velocity and the Y axis is + or -
+                            // And then go from there!
                             -(axis.angle_between(-glam::Vec3::Y)).signum()
                                 * velocity.length() as f32,
                         )
--- a/chariot-client/src/graphics.rs
+++ b/chariot-client/src/graphics.rs
@@ -258,26 +258,43 @@ impl GraphicsManager {
             .expect("Trying to update player location when transform does not exist");
         *player_transform = Transform::from_entity_location(&location, player_transform.scale);
 
-        // if this player is the main player, update the camera too (based on velocity)
-        if player_entity == self.camera_entity && *velocity != DVec3::ZERO {
+        if player_entity == self.camera_entity {
             if let Some(camera) = self.world.get_mut::<Camera>(self.camera_entity) {
-                // first we have to compensate for the rotation of the chair model
-                let rotation_angle = location.unit_steer_direction.angle_between(DVec3::X);
-                // next, we add the angle of the direction of the velocity
-                let velocity_angle =
-                    DVec3::new(velocity.x, 0.0, velocity.z).angle_between(DVec3::X);
+                match self.player_choices[player_num]
+                    .as_ref()
+                    .unwrap()
+                    .clone()
+                    .chair
+                    .cam()
+                {
+                    chariot_core::player::choices::CameraType::FaceForwards => {
+                        camera.orbit_angle = Vec2::new(0.0, -0.3).lerp(camera.orbit_angle, 0.5);
+                    }
+                    chariot_core::player::choices::CameraType::FaceVelocity => {
+                        if *velocity != DVec3::ZERO {
+                            // first we have to compensate for the rotation of the chair model
+                            let rotation_angle =
+                                location.unit_steer_direction.angle_between(DVec3::X);
+                            // next, we add the angle of the direction of the velocity
+                            let velocity_angle =
+                                DVec3::new(velocity.x, 0.0, velocity.z).angle_between(DVec3::X);
 
-                // there's actually some magic trig cancellations happening here that simplify this calculation
-                let mut orbit_yaw = PI + location.unit_steer_direction.z.signum() * rotation_angle;
+                            // there's actually some magic trig cancellations happening here that simplify this calculation
+                            let mut orbit_yaw = velocity.z.signum() * velocity_angle
+                                - location.unit_steer_direction.z.signum() * rotation_angle;
 
-                // if the yaw change would be bigger than PI, wrap back around
-                let yaw_difference = orbit_yaw - camera.orbit_angle.x as f64;
-                if yaw_difference.abs() > PI {
-                    orbit_yaw += yaw_difference.signum() * 2.0 * PI;
-                }
+                            // if the yaw change would be bigger than PI, wrap back around
+                            let yaw_difference = orbit_yaw - camera.orbit_angle.x as f64;
+                            if yaw_difference.abs() > PI {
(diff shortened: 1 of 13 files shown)


KNOXDEV:
rolling beanbag fluid physics when

TheCrossBoy:
New buttons! Left arrow and right arrow cycle through the chair choices.

TheCrossBoy:
Also: there will likely be some weird file name shenanegains with this, because I want the names to be clear after this PR. Will make sure the drive is updated when it becomes relevant.

trulyronak:
swivel - i fucking love this chair
recliner sofa thing - turning with this is kinda hard ngl, the camera works near perfect with swivel but i don't really find it doing what i want it to with this one (just feels awk turning)
ergonomic - it feels like the swivel but without the cool rotating, but i feel like it's not different enough from swivel visually (something to do on the art side)
beanbag - this is pure chaos i love it but i think we should make the turning better so it's not purely a meme

chariot-client/src/graphics.rs
```diff
@@ -256,30 +257,70 @@ impl GraphicsManager {
...
-
-                // if the yaw change would be bigger than PI, wrap back around
-                let yaw_difference = orbit_yaw - camera.orbit_angle.x as f64;
-                if yaw_difference.abs() > PI {
-                    orbit_yaw += yaw_difference.signum() * 2.0 * PI;
+        let new_player_transform =
+            Transform::from_entity_location(&location, player_transform.scale);
+        *player_transform = new_player_transform;
+
+        if velocity.length() > 0.0 {
```
trulyronak, line 264: trying to understand some of the graphics code: what's the purpose of this if statement?

chariot-client/src/graphics.rs
```diff
@@ -256,30 +257,70 @@ impl GraphicsManager {
...
+                {
+                    let axis = velocity.as_vec3().cross(glam::Vec3::Y).normalize();
+                    drawable.modifiers.rotation = Some(
+                        glam::Quat::from_axis_angle(
+                            axis,
+                            -(axis.angle_between(-glam::Vec3::Y)).signum()
+                                * velocity.length() as f32,
+                        )
+                        .normalize()
+                        .mul_quat(drawable.modifiers.rotation.unwrap_or_default()),
```
trulyronak, line 281: this feels arbitrary to me so i'm just confused by what this does

chariot-client/src/graphics.rs
```diff
@@ -256,30 +257,70 @@ impl GraphicsManager {
...
-                camera.orbit_angle =
-                    Vec2::new(orbit_yaw as f32, -0.3).lerp(camera.orbit_angle, 0.5);
+        if player_entity == self.camera_entity {
+            if let Some(camera) = self.world.get_mut::<Camera>(self.camera_entity) {
+                match self.player_choices[player_num]
+                    .as_ref()
+                    .unwrap()
+                    .clone()
+                    .chair
+                    .cam()
```
trulyronak, line 294: neat: loving the methods you've added to the chair enum

chariot-client/src/graphics.rs
```diff
@@ -370,17 +411,40 @@ impl GraphicsManager {
...
+                    if drawable.modifiers.absolute_angle {
+                        acc_model = *acc
+                            * Transform {
+                                translation: cur_transform.translation,
+                                rotation: glam::Quat::IDENTITY,
+                                scale: cur_transform.scale,
+                            }
+                            .to_mat4();
+                    } else if let Some(rotation) = drawable.modifiers.rotation {
+                        // println!("d {}", rotation);
```
trulyronak, line 437: nit: unneeded comment

chariot-client/src/scenegraph/components.rs
```diff
@@ -65,6 +65,12 @@ impl Default for Transform {
     }
 }
 
+#[derive(Default, Clone, Copy)]
+pub struct Modifiers {
```
trulyronak, line 69: do these modifers exist for any Transform in the screengraph? Also I'm not fully understanding why absolute angle is required and rotation isn't

chariot-core/src/player/choices.rs
```diff
@@ -1,28 +1,180 @@
...
+            chair: Chair::Swivel,
+            map: Track::Track,
+            ready: false,
+        }
+    }
+}
+
+lazy_static! {
+    static ref DEFAULT_CHAIR: HashMap<String, f64> = [
+        ("gravity_coefficient", GLOBAL_CONFIG.gravity_coefficient),
```
trulyronak, line 28: MOON GRAVITY WHEN

chariot-core/src/player/choices.rs
```diff
@@ -1,28 +1,180 @@
...
+        &*DEFAULT_CHAIR
+    }
+
+    pub fn scale(&self) -> DVec3 {
+        match self {
+            _ => DVec3::new(1.0, 2.0, 1.0),
+        }
+    }
+
+    pub fn stat(&self, stat_name: &str) -> f64 {
```
trulyronak, line 81: can we make `stat_name` an enum as well?

chariot-core/src/player/choices.rs
```diff
@@ -1,28 +1,180 @@
...
+                    GLOBAL_CONFIG.rolling_resistance_coefficient * 0.5
+                }
+
+                "drag_coefficient" => GLOBAL_CONFIG.drag_coefficient * 0.02,
+                _ => *Chair::default_stats().get(stat_name).unwrap(),
+            },
+            Chair::Recliner => match stat_name {
+                // Make our turn very hefty
+                "car_spin" => GLOBAL_CONFIG.car_spin * 0.06,
+                "max_car_spin" => GLOBAL_CONFIG.max_car_spin * 0.7,
```
trulyronak, line 95: 🌶️ can't wait to break this

chariot-core/src/player/choices.rs
```diff
@@ -1,28 +1,180 @@
...
+                "drag_coeffecient" => GLOBAL_CONFIG.drag_coefficient * 1.5,
+                // And we have a decent break
+                "car_brake" => GLOBAL_CONFIG.car_brake * 15.0,
+                _ => *Chair::default_stats().get(stat_name).unwrap(),
+            },
+            Chair::Beanbag => match stat_name {
+                "max_car_speed" => GLOBAL_CONFIG.max_car_speed * 1.8,
+                "car_accelerator" => GLOBAL_CONFIG.car_accelerator * 0.33,
+                // A L L G A S N O B R A K E S
+                "car_brake" => 0.0,
```
trulyronak, line 130: 🥵

chariot-core/src/player/choices.rs
```diff
@@ -1,28 +1,180 @@
...
+                    GLOBAL_CONFIG.rolling_resistance_coefficient * 0.1
+                }
+                "mass" => 1.0,
+                _ => *Chair::default_stats().get(stat_name).unwrap(),
+            },
+            Chair::Folding => match stat_name {
+                _ => unimplemented!(
+                    "I don't have a lot of ideas for this one currently. May add later!"
+                ),
+                //_ => *Chair::default_stats().get(stat_name).unwrap(),
```
trulyronak, line 143: nit: commented code

chariot-core/src/player/choices.rs
```diff
@@ -1,28 +1,180 @@
...
+                "drag_coefficient" => GLOBAL_CONFIG.drag_coefficient * 0.001,
+                "rolling_resistance_coefficient" => {
+                    GLOBAL_CONFIG.rolling_resistance_coefficient * 0.1
+                }
+                "mass" => 1.0,
+                _ => *Chair::default_stats().get(stat_name).unwrap(),
+            },
+            Chair::Folding => match stat_name {
+                _ => unimplemented!(
+                    "I don't have a lot of ideas for this one currently. May add later!"
```
trulyronak, line 141: ideas: make it like the og things you did by accident and have the chair rotate around itself entirely over and over again

trulyronak (approved):
overall quite hype, i had some questions re graphics stuff that I kinda want explanations for, but if its trivial for someone who knows graphics I guess dw about it too much

requested changes re making some stuff an enum + removing some comments, but overall p good! gonna approve + i think we can easily tweak physics later as well.

chariot-client/src/graphics.rs
alistairgray42: there's a divide-by-zero condition that happens if velocity is zero

trulyronak: ah that makes sense

trulyronak: follow up: why would checking the length of the vector see if its 0?

chariot-core/src/player/choices.rs
alistairgray42: IT'S LEGIT SUCH AN EASY AUDIENCE INTERACTION TO CODE (but only really makes sense if we get ramps oop)

chariot-core/src/settings.rs
```diff
@@ -71,15 +73,17 @@ impl Settings {
             // the rotational analogue of friction: each time step in free
             // rotation, what proportion of angular velocity should be
             // conserved?
-            .set_default("rotation_reduction_coefficient", 0.80)?
+            .set_default("rotation_reduction_coefficient", 0.40)?
             // Coefficient to control how forceful player-controlled acceleration is
-            .set_default("car_accelerator", 0.8)?
+            .set_default("car_accelerator", 0.8 / 15.0)?
```
alistairgray42, line 78: any reason not to just make this an actual constant?

chariot-core/src/player/choices.rs
trulyronak: so im hearing RAMPS SOON

chariot-core/src/settings.rs
trulyronak: question: why 15.0?

chariot-client/src/graphics.rs
TheCrossBoy: You can also think of it conceptually - if the length of the vector is zero, how much is the bag rolling?

Since it's speed is zero, it isn't moving at all, so it shouldn't be moved at all!

TheCrossBoy: Oh and as to why we are getting the length, the velocity vector tells us two things:

- which direction we are moving

- how fast we are moving in that direction 

The direction is the normalized vector, and the speed is the length of it (aka the magnitude)

chariot-core/src/player/choices.rs
TheCrossBoy: The only reason I didn't is because it doesn't actually help at all. You still need to map string => something in global_config with a match, so it didn't seem particularly helpful. But if we want it I can

chariot-core/src/player/choices.rs
TheCrossBoy: This was literally my original idea but I thought you guys would think it was too - insane but now I'm gonna do it LMAO

chariot-client/src/graphics.rs
trulyronak: gotcha, i think i was thining of velocity vector as a [0,0,0] kind of thing

chariot-core/src/player/choices.rs
trulyronak: i personally support making it an enum because I don't want to allow any version of a typo breaking something; the match statements imo shouldn't be working off of direct strings here / if we reference a stat it isn't great to have 'magic numbers' or in this case a magic string.

potentially also we might extend on stats so that might be nice

chariot-core/src/player/choices.rs
trulyronak: HYPE

chariot-client/src/graphics.rs
TheCrossBoy: That is what it is! It's a vector with 3 positions (x,y,z) that tells you where you're going/how fast.

chariot-client/src/graphics.rs
TheCrossBoy: Sure, here's a description/overview (I can add comments too):

In order to get the rotation, we need an *axis* of rotation (what we spin around) and an *angle* to spin (how far around the axis we spin). We essentially want to spin "towards" the velocity: the top of the bag should move towards our velocity. 

So, for our axis of rotation, we want it to be perpendicular to the velocity's direction. You can see this if you have your finger pointing forwards: if you try to move it to point down, the axis of rotation is basically the "right-left" line. 

But there are *two* lines perpendicular to the velocity: one pointed up-down and one left-right. If we take the cross product with the Y axis (up down), that basically gets us the other one, aka the right-left one we want. 

Now, we just need the amount to spin! We can just set this to be the length of the velocity, because it's *close enough* to what it would actually be. But there's a problem: length is always positive, but sometimes we spin in the opposite direction.

Remember, we always want to spin the velocity "down". So, if we get the angle between our velocity and the Y axis (which is either 90° or -90°), and take the sign of that, we will get which direction we want to go to move the velocity's direction down!

So put that all together and you get this. 

Lmk if you need more or are still confused!

chariot-client/src/scenegraph/components.rs
TheCrossBoy: Yes, they do. The only reason that it's an option for this is to prevent unnecessary calculations. 

If the absolute angle is set to true, it will prevent angle rotations, and do an extra calculation to do that. 

If there is a rotation we need to apply, then we do an extra calculation to apply that. 

We could just have it such that it just has a default rotation of 0° (aka nothing), but then we'd have to apply that on *every* mesh, which is a bunch of extra steps as opposed to just skipping it when not necessary.

chariot-client/src/graphics.rs
trulyronak: this is fantastic, i would personally like to see some of this reflected in comment form in the code, but up to your discretion

chariot-core/src/settings.rs
TheCrossBoy: Why it's not a constant: .8/15 isn't an even number, so it was just cleaner to keep it like this.

Why 15: it's just the coincidental value I happened to find as feeling the best. Originally, I just kept it all the same, then was tweaking chairs individually, but noticed that everything was needing a much lower acceleration, so I just went ahead and basically "moved" the custom shifting I was doing for the swivel chair into the default, and adjusted the rest accordingly. 

I would be fine changing as needed though! Realistically these values don't have an exact "meaning" anyways, since having 1.1 vs 4.1 doesn't really have a real meaning (irl it'd be miles per hour^2, which are both units we know, but there's no "real" unit here we can use, so the only use is in relation to another value, i.e. a higher or lower acceleration), so it's probably not that important to have a specific value for it

TheCrossBoy:
Resolved most comments. Tomorrow:

- [ ] Mess with Recliner cam
- [ ] Try something with Ergonomic acceleration, or nerf swivel in some way
- [ ] Give beanbag more control
- [ ] Make stat enum
- [ ] Implement plastic chair

PR #62, opened 2022-05-20
Network Resilience

KNOXDEV:
Two parts to this PR:

- [ ] on the client, if we disconnect from the server, rather than panic, keep attempting connections with exponential backoff and only panic if we can't connect for a full 30 seconds. If the reconnect is successful, pick up the game where we left off
- [ ] on the server, if a client disconnects, freeze the game loop until they have reconnected. If they do not reconnect after 30 seconds, die

### DIFF
--- a/chariot-client/src/graphics.rs
+++ b/chariot-client/src/graphics.rs
@@ -147,7 +147,7 @@ impl GraphicsManager {
 
         let mut test_string = StringDrawable::new("ArialMT", 18.0);
         test_string.set(
-            "chariot - 0.6.9",
+            "chariot - 0.6.11",
             Vec2::new(0.005, 0.027),
             &renderer,
             &mut resources,


KNOXDEV:
Basically this addresses the most common cases, but not all of the cases I originally sought out to handle. That's alright though, I also squashed all client and server sided errors while I was in there.

chariot-client/src/audio/thread/mod.rs
```diff
@@ -1,8 +1,10 @@
+#![allow(dead_code)]
```
TheCrossBoy, line 1: Unless there's a lot of it here, I think it's better for us to allow the dead code one by one (with a comment explaining why its being kept), rather than as a full file

chariot-client/src/audio/mod.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(dead_code)]
```
TheCrossBoy, line 1: Some comment here

chariot-client/src/scenegraph/particle_system.rs
```diff
@@ -1,15 +1,11 @@
-use std::ops::Bound;
-
+#![allow(dead_code)]
```
TheCrossBoy, line 1: Same as above

chariot-client/src/ui/ui_region.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(dead_code)]
```
TheCrossBoy, line 1: Same

trulyronak:
agree with @alistairgray42 's comment, let's hold off on merging this until #81 is merged so @nickwn doesn't have to deal with more merge conflicts

chariot-core/src/networking/connection.rs
```diff
@@ -63,11 +63,11 @@ impl<T: Packet, V: Packet> Connection<T, V> {
             // if we parsed a packet size, let's go ahead and read that amount,
             // this time blocking until we've parsed the entire thing
             self.set_blocking();
-            let packet =
+            if let Ok(packet) =
                 T::parse_packet(&mut Read::by_ref(&mut self.tcp_stream).take(packet_size as u64))
-                    .expect("Failed to deserialize packet");
-
-            self.incoming_packets.push_back(packet);
+            {
```
TheCrossBoy, line 68: We might wanna print some kind of error if we do fail to deserialize, just so we can know what's going on

TheCrossBoy (approved):
Assuming it all works, then looks great. Tested locally, only note is it didn't seem to work locally, but I attempted to test this by pressing escape to exit out of the client, but that just forcibly closed the connection, which still crashes the server

chariot-client/src/audio/thread/mod.rs
TheCrossBoy: Btw assume these are all nits ofc i forgot to add that

PR #63, opened 2022-05-21
In Game UI

trulyronak:
The goal of this PR is to make it so you can tell what's going on during a game while looking at just the window
- [x] should be able to see what place you're in
- [x] should be able to see what audience interaction events are happening right now
- [x] should be able to see how long a vote's effects will last in the game

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,15 +1,11 @@
 {
 	// See https://go.microsoft.com/fwlink/?LinkId=827846 to learn about workspace recommendations.
 	// Extension identifier format: ${publisher}.${name}. Example: vscode.csharp
-
 	// List of extensions which should be recommended for users of this workspace.
 	"recommendations": [
-		"matklad.rust-analyzer",
+		"rust-lang.rust-analyzer",
 		"PolyMeilex.WGSL"
-		
 	],
 	// List of extensions recommended by VS Code that should not be recommended for users of this workspace.
-	"unwantedRecommendations": [
-		
-	]
+	"unwantedRecommendations": []
 }
--- a/chariot-client/src/application.rs
+++ b/chariot-client/src/application.rs
@@ -11,8 +11,10 @@ use chariot_core::player::player_inputs::{EngineStatus, InputEvent, RotationStat
 use chariot_core::GLOBAL_CONFIG;
 
 use crate::game::GameClient;
-use crate::graphics::{register_passes, AnnouncementState, GraphicsManager, UIState};
+use crate::graphics::{register_passes, GraphicsManager};
+
 use crate::ui::ui_region::UIRegion;
+use crate::ui_state::{AnnouncementState, UIState};
 
 pub struct Application {
     pub graphics: GraphicsManager,
@@ -162,8 +164,6 @@ impl Application {
                     question,
                     time_until_vote_end,
                 } => {
-                    println!("The voting has started");
-                    // println!("{:?}", self.graphics.ui);
                     let vote_end_time = Instant::now() + time_until_vote_end;
                     self.graphics.make_announcement(
                         "The audience is deciding your fate",
--- a/chariot-client/src/application.rs
+++ b/chariot-client/src/application.rs
@@ -209,6 +209,7 @@ impl Application {
                 }
             }
         }
+        self.graphics.update_minimap();
     }
 
     // Input configuration
--- a/chariot-client/src/application.rs
+++ b/chariot-client/src/application.rs
@@ -73,7 +73,8 @@ impl Application {
                 );
                 if effect_end_time < Instant::now() {
                     self.game.announcement_state = AnnouncementState::None;
-                    self.graphics.make_announcement("", "");
+                    self.graphics.game_announcement_title.should_draw = false;
+                    self.graphics.game_announcement_subtitle.should_draw = false;
                 }
             }
             AnnouncementState::GeneralAnnouncement { title, subtitle } => {}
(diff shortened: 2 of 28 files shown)


.vscode/extensions.json
```diff
@@ -1,15 +1,11 @@
 {
```
alistairgray42, line 1: (not actually sure) is it usual to gitignore .vscode? I'd think at least `settings.json` should be gitignored

chariot-client/src/application.rs
```diff
@@ -43,6 +47,43 @@ impl Application {
     }
 
     pub fn render(&mut self) {
+        if let UIState::InGameHUD {
```
alistairgray42, line 50: All this logic should go at least one level down into `GraphicsManager::render`, and almost surely at least one more level down into another function that's called from there (since we're worried about too much cruft accumulating in `graphics.rs`)

chariot-client/src/application.rs
```diff
@@ -43,6 +47,43 @@ impl Application {
...
+            announcement_state, ..
+        } = &self.graphics.ui
+        {
+            match announcement_state {
+                AnnouncementState::VotingInProgress {
+                    prompt: _,
```
alistairgray42, line 56: Nit: use `..` whenever possible (for `VoteActiveTime` as well)

chariot-client/src/application.rs
```diff
@@ -99,17 +140,74 @@ impl Application {
                 }
                 ClientBoundPacket::PlacementUpdate(position) => {
                     println!("I am now placed {}!", position);
+                    if let UIState::InGameHUD {
```
alistairgray42, line 143: This logic should also probably go in GraphicsManager for delegation / cleanliness

chariot-client/src/drawable/string.rs
```diff
@@ -23,10 +34,18 @@ impl StringDrawable {
...
         content: &str,
-        mut screen_pos: Vec2,
         renderer: &Renderer,
         resource_manager: &mut ResourceManager,
     ) {
+        let mut screen_pos = self.screen_position.clone();
+
+        if self.center_text {
+            // to center the screen position:
+            let letter_width = 11.0 / 741.0;
```
alistairgray42, line 44: where do these numbers come from?

chariot-client/src/application.rs
```diff
@@ -99,17 +140,74 @@ impl Application {
...
+                    self.graphics.make_announcement(
+                        "The audience is deciding your fate",
+                        format!(
+                            "They decide in {} seconds",
+                            (vote_end_time - Instant::now()).as_secs()
+                        )
+                        .as_str(),
                     );
+
+                    if let UIState::InGameHUD {
```
alistairgray42, line 177: I suggest replacing this `if let` pattern, which shows up a bunch and is pretty inelegant, with this method, throughout:

```
impl UIState {
    pub fn maybe_set_announcement_state(&mut self, new_announcement_state: AnnouncementState) {
        if let UIState::InGameHUD {
            announcement_state, ..
        } = self
        {
            *announcement_state = new_announcement_state;
        }
    }
}

```

alistairgray42 (changes requested):
I can't actually run this on my own computer because of some font nonsense, so I will assume this all works for now?

I think announcements should have timeouts - preferably fade after a few seconds, but just disappearing is good too. I'd also like to see more organizational modularization

.vscode/extensions.json
trulyronak: the file wasn't gitignored for workspace recs, these changes seem *okay* to go through for now

chariot-client/src/drawable/string.rs
trulyronak: these numbers are really bad approximations of the screen size; I'm using this temporarily until we add text centering (didn't want to add that for this pr since it'd get this too out of scope; this is mvp)

chariot-client/src/application.rs
alistairgray42: (or something like this ofc)

trulyronak:
To get the font — go to https://fonts.google.com/specimen/Press+Start+2P and hit "Download All"

TheCrossBoy (commented):
Overall, neat stuff! UI is def something we should talk about at the meeting today though, since it would be nice to get somethign well organized agreed upon.

chariot-client/src/drawable/string.rs
```diff
@@ -8,13 +8,24 @@ use glam::Vec2;
 pub struct StringDrawable {
     ui_drawable: UIDrawable,
     glyph_cache: GlyphCache,
+    pub should_draw: bool,
+    pub screen_position: Vec2,
+    pub center_text: bool,
```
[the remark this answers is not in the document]
trulyronak: agree i like the idea of an alignment enum instead of center_Text

chariot-client/src/resources/mod.rs
```diff
@@ -170,10 +170,12 @@ impl ResourceManager {
             .map(|n| (n, glam::Mat4::IDENTITY))
             .collect::<VecDeque<(gltf::Node, glam::Mat4)>>();
 
+        let mut n = 0;
         // Probably better to do this recursively but i didn't wanna change stuff like crazy, not that it really matters since this is just loading anyways
         while let Some((node, parent_transform)) = queue.pop_front() {
             println!("Processing node '{}'", node.name().unwrap_or("<unnamed>"));
-
+            println!("#{} out of #{}", n, queue.capacity());
```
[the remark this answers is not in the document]
trulyronak: the first one was there from before, i guess it explains which node it is, the second one just shows the progress (so i can see how much longer I have to wait)

alistairgray42 (approved):

TheCrossBoy (approved):

PR #64, opened 2022-05-22
particle effects

nickwn:
still a few things I want to change so it's a draft


chariot-client/src/drawable/particle.rs
```diff
@@ -0,0 +1,81 @@
+use super::Drawable;
+use super::TransformUniform;
+use crate::renderer::render_job;
+use crate::renderer::Renderer;
+use crate::resources::{Handle, MaterialHandle, ResourceManager, StaticMeshHandle};
+
+// TODO: merge this with the forward pass and make this write depth. I got it to work but it was weird since on certain frames
+// the particles wouldn't show up and the outline effect was being applied to the quads
```
TheCrossBoy, line 8: Is this going to be done on this PR, is it already done, or can be it be done in another PR?

chariot-client/src/scenegraph/mod.rs
```diff
@@ -8,6 +8,9 @@ use components::*;
 use crate::resources::{accum_bounds, new_bounds, Bounds};
 
 pub mod components;
+pub mod particle_system;
+
+pub use particle_system::*;
```
TheCrossBoy, line 13: Should this be here? Wouldn't this get rid of all scoping on particle system things? Or do we want it that way

chariot-client/src/scenegraph/particle_system.rs
```diff
@@ -0,0 +1,310 @@
...
+            lifetime: 0.0,
+            rotation: ParticleRotation::Billboard,
+            gravity: 0.0,
+        }
+    }
+}
+
+// Because I made the good/bad decision to incorporate particles into the scene graph,
+// Particle systems have an ID worked into the type so they can have their own component types.
+// Ex. ParticleSystem<0> would operate on all the BillboardParticle<0> components in a world.
```
TheCrossBoy, line 102: I'm slightly confused - why would this be required for it to be part of the scene graph?

chariot-client/src/scenegraph/particle_system.rs
```diff
@@ -0,0 +1,310 @@
...
+
+    // Calculates the local transform for a particle
+    pub fn calc_particle_model(
+        &self,
+        world: &World,
+        entity: Entity,
+        view: glam::Mat4,
+    ) -> Option<glam::Mat4> {
+        // TODO (maybe): not the fastest thing in the world
+        let (view_scale, view_rot, view_trans) = view.inverse().to_scale_rotation_translation();
```
TheCrossBoy, line 292: honestly cargo run --release is more op than most optimization we can realistically do in a short time frame, but if it becomes an issue then we can figure something out that that point imo

TheCrossBoy (commented):
Looks good! If you could just make sure to merge main and commit your resource files to git (since main removes that ignored folder atm) it'll be good to go, assuming my comments are fine too

TheCrossBoy (approved):
I meant to approve pls see previous review for relevant stuff

chariot-client/src/drawable/particle.rs
nickwn: I'll leave it for later. I tried to do it in this PR but things got a little tricky and I think it's also related to some problems I need to fix with the UI.

chariot-client/src/scenegraph/particle_system.rs
nickwn: I wanted a way to iterate over all the particles from system 1, apply the system 1 specific update to those particles, then do the same for system 2 which might have slightly different update parameters (ex different gravity). However, if there were a single particle component class (ie. a number wasn't built into the type), then I would only be able to iterate over all particles from both systems, not each individually.

chariot-client/src/scenegraph/mod.rs
nickwn: got rid of it, it was me being lazy

PR #66, opened 2022-05-22
ramps c:

alistairgray42:
:ron_paul_its_happening_dot_gif:

capabilities:
- you can go up ramps!
- you get a little speed boost while going up em so it's not boringly realistic
- running into the back / intended non-slopey part of the ramp is like running into any other solid terrain
- gravity and mass are tuned so that you can glide for a while
- ramp gets loaded in from blender models

limitations (at the moment):
- a lot of this code is just horrendously ugly ngl
- camera does not make as much sense as it could when going off the ramp


KNOXDEV:
I'll try this out later today, but I won't approve until the tests are fixed FYI

chariot-core/src/player/choices.rs
```diff
@@ -43,7 +43,7 @@ lazy_static! {
         (Stat::CarSpin, GLOBAL_CONFIG.car_spin),
         (Stat::MaxCarSpeed, GLOBAL_CONFIG.max_car_speed),
         (Stat::MaxCarSpin, GLOBAL_CONFIG.max_car_spin),
-        (Stat::Mass, 10.0),
+        (Stat::Mass, 0.01),
```
trulyronak, line 46: not even god understands how this number relates to the real force of gravity

trulyronak: also, since we have a global config stat, we should reference it
```suggestion
        (Stat::Mass, GLOBAL_CONFIG.gravity_coefficient),
```

chariot-server/src/game/map.rs
```diff
@@ -102,7 +105,7 @@ impl Map {
 
         // Probably better to do this recursively but i didn't wanna change stuff like crazy, not that it really matters since this is just loading anyways
         while let Some((node, parent_transform)) = queue.pop_front() {
-            //println!("Processing node '{}'", node.name().unwrap_or("<unnamed>"));
+            println!("Processing node '{}'", node.name().unwrap_or("<unnamed>"));
```
trulyronak, line 108: note: we should prolly remove these print statements at some point

chariot-server/src/game/map.rs
```diff
@@ -175,6 +178,32 @@ impl Map {
                                     mesh.name().unwrap_or("<unnamed>")
                                 );
                                 colliders.push(mesh_bounds);
+                            } else if purpose == "ramp" {
+                                let incline_direction: String = mesh_data
+                                    .get("incline_direction")
+                                    .expect("Ramps should have an incline direction!")
```
trulyronak, line 184: question - this might be in documentation later that I will read (if so this comment will get deleted) — is the incline direction something that is set by blender? Can this be modified?

chariot-server/src/physics/player_entity.rs
```diff
@@ -140,8 +117,22 @@ impl PlayerEntity {
         potential_colliders: Vec<&PlayerEntity>,
         potential_terrain: Vec<BoundingBox>,
         potential_triggers: impl Iterator<Item = &'a mut dyn TriggerEntity>,
+        ramp_collision_result: Option<&RampCollisionResult>,
     ) -> PlayerEntity {
-        let self_forces = self.sum_of_self_forces();
+        let minimum_player_height =
+            if ramp_collision_result.is_some() && ramp_collision_result.unwrap().can_get_on {
```
trulyronak, line 123: nit: since you're unwrapping this twice, I like to take advantage of rust's renaming and just do
```rust
let ramp_collision_result = ramp_collision_result.unwrap()
```
so we don't need to unwrap multiple times later

chariot-server/src/physics/player_entity.rs
```diff
@@ -186,6 +178,21 @@ impl PlayerEntity {
...
+            && ramp_collision_result.unwrap().can_get_on
+            && !self.is_aerial(ramp_collision_result)
+        {
+            let ramp_incline = ramp_collision_result
+                .unwrap()
+                .ramp
+                .get_incline_vector()
+                .normalize();
+
+            // vroom vroom vroom
```
trulyronak, line 192: 🚀🚗🚀🚗

chariot-server/src/physics/player_entity.rs
```diff
@@ -249,24 +270,37 @@ impl PlayerEntity {
         return new_player;
     }
 
-    fn is_aerial(&self) -> bool {
-        return self.entity_location.position[1]
-            > self.size[1]
-                + get_height_at_coordinates(
+    fn is_aerial(&self, ramp_collision_result: Option<&RampCollisionResult>) -> bool {
```
trulyronak, line 273: nit: handling a `None` `ramp_collision_result` should be in the caller, not in this method (this method shouldn't have to worry about `ramp_collision_result` being `None`

chariot-server/src/physics/player_entity.rs
```diff
@@ -249,24 +270,37 @@ impl PlayerEntity {
...
+        } else {
+            0.0
+        };
+
+        self.entity_location.position[1] - 1.0 > ground_level
     }
 
-    fn sum_of_self_forces(&self) -> DVec3 {
-        let air_forces = self.gravitational_force_on_object()
+    fn sum_of_self_forces(&self, ramp_collision_result: Option<&RampCollisionResult>) -> DVec3 {
```
trulyronak, line 289: nit: same comment re `ramp_collision_result`

chariot-server/src/physics/ramp.rs
```diff
@@ -0,0 +1,240 @@
+use glam::{DQuat, DVec2, DVec3};
+
+use super::{bounding_box::BoundingBox, player_entity::PlayerEntity};
+
+#[derive(Clone, Copy, Debug)]
+pub struct Ramp {
+    // [[min_x, max_x]; [min_z, max_z]]
+    pub footprint: [[f64; 2]; 2],
```
trulyronak, line 8: suggestion: would recommend using [type aliases](https://doc.rust-lang.org/reference/items/type-aliases.html) throughout this file, especially to help understand the numbers for all the physicsy things

chariot-server/src/physics/tests.rs
```diff
@@ -69,87 +69,17 @@ fn get_origin_cube() -> PlayerEntity {
...
-    let expected_velocity = prev_velocity
-        + (neg_prev_velocity / neg_prev_velocity.length()) * GLOBAL_CONFIG.car_brake
-        + neg_prev_velocity * GLOBAL_CONFIG.drag_coefficient * (5.0 as f64).sqrt()
-        + neg_prev_velocity * GLOBAL_CONFIG.rolling_resistance_coefficient;
-    assert!(props.velocity.abs_diff_eq(
-        expected_velocity.normalize() * GLOBAL_CONFIG.max_car_speed,
-        0.001
-    ));
-}
-
```
trulyronak, line 141: https://www.youtube.com/watch?v=rR4n-0KYeKQ

trulyronak (approved):
generally looks good, left mostly comments on rust style + some ways to potentially help w/ maintainability

chariot-server/src/game/map.rs
alistairgray42: this is a custom property that we ourselves set in blender, easily modifiable

chariot-server/src/physics/player_entity.rs
alistairgray42: yeah, i'm using `None` semantically here - bad choice, I'll add an enum to make this clearer

chariot-server/src/physics/tests.rs
alistairgray42: so true https://discord.com/channels/958422216498896956/960773822385106984/978700086303404062

chariot-server/src/game/map.rs
alistairgray42: If I am not for myself, who will be for me? And when I am for myself, what am 'I'? And if not now, when?

chariot-server/src/physics/ramp.rs
```diff
@@ -0,0 +1,246 @@
...
+        // the ramp, or in the strip leading out from the ramp face
+        if x >= ramp_min_x && x <= ramp_max_x && z >= ramp_min_z && z <= ramp_max_z {
+            return ramp.get_height_at_coordinates(x, z) - (self.entity_location.position.y - 1.0)
+                < 1.0;
+        }
+
+        // to enter the ramp, you must be within the strip leading out from the
+        // low side of the ramp by 1 unit, and have velocity that's pointed along the
+        // incline direction
+        if ramp.incline_direction == DVec2::X {
```
trulyronak, line 143: Curious: could this become an enum?

alistairgray42: In the case where we only have axis-orthogonal ramps, yes; I designed most of this code (not this one method cause it would be annoying) to allow for non-axis-orthogonal ramps, however, and in that case it can't be an enum.

PR #68, opened 2022-05-23
add controller support

TheCrossBoy:
its basically already done how hard can it be

### DIFF
--- a/chariot-client/src/application.rs
+++ b/chariot-client/src/application.rs
@@ -331,13 +331,9 @@ impl Application {
                     /***** CONTROLS *****/
                     // Accelerator
                     Button::RightTrigger2 if value > 0.0 => {
-                        println!("New gamepad event: {:?}", event);
                         Some(InputEvent::Engine(EngineStatus::Accelerating(value)))
                     }
-                    Button::RightTrigger2 => {
-                        println!("New gamepad event: {:?}", event);
-                        Some(InputEvent::Engine(EngineStatus::Neutral))
-                    }
+                    Button::RightTrigger2 => Some(InputEvent::Engine(EngineStatus::Neutral)),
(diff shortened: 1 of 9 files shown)


chariot-client/src/application.rs
```diff
@@ -312,4 +315,126 @@ impl Application {
...
+                        Some(InputEvent::Engine(EngineStatus::Accelerating(value)))
+                    }
+                    Button::RightTrigger2 => {
+                        println!("New gamepad event: {:?}", event);
+                        Some(InputEvent::Engine(EngineStatus::Neutral))
+                    }
+                    // Brake
+                    Button::LeftTrigger2 => None,
+
+                    /***** MENU *****/ // TODO: this is temporary. we need a real way to handle menu input w controllers
```
trulyronak, line 340: dear god i completely forgot about this LOL

chariot-client/src/application.rs
```diff
@@ -312,4 +315,126 @@ impl Application {
...
+                    }
+                    _ => None,
+                }
+            }
+            // Axis (value: [-1, 1])
+            Err((axis, value)) => match axis {
+                /***** MOVEMENT *****/
+                // Turn right
+                Axis::LeftStickX if value > 0.0 => {
+                    println!("New gamepad event: {:?}", event);
```
trulyronak: can you omit the printlns for gamepad events if we do something with it already

trulyronak (approved):
lgtm! just left a comment on removing unneeded printlns

chariot-client/src/application.rs
TheCrossBoy: I think for now we can just like... ignore it lol

We can add it later if we want to lmfao

PR #69, opened 2022-05-24
font rendering polish steps

KNOXDEV:
Still a WIP. Here's what we're gonna tackle:

- [ ] glyphs own their textures and resource manager owns glyph caches
- [ ] formal API for string alignment
- [ ] address kerning issues
- [ ] add api to load fonts from resource files in addition to current system fonts
- [ ] address transparency issues


KNOXDEV:
lucky PR number

chariot-client/src/drawable/string.rs
```diff
@@ -1,49 +1,86 @@
...
 
-impl StringDrawable {
-    pub fn new(font_name: &str, point_size: f32, screen_position: Vec2) -> StringDrawable {
-        StringDrawable {
-            ui_drawable: UIDrawable { layers: vec![] },
-            glyph_cache: GlyphCache::new(font_name, point_size),
-            screen_position,
-            center_text: false,
+// builder-pattern structure for creating UIDrawables that represent a rendered string
+impl StringBuilder {
```
trulyronak, line 25: Whats your take on organizing this stuff into the ui mod? Something like "UIString"

Primarily because all the text stuff makes sense to be in the ui mod

chariot-client/src/renderer/render_job.rs
```diff
@@ -256,6 +256,12 @@ impl<'a> RenderJob<'a> {
             .into_iter()
             .zip(job_root_ids.into_iter())
             .collect();
+
+        // if this render graph is empty, this is a no-op function
+        if stack.is_empty() {
+            return;
+        }
```
trulyronak, line 263: Is this so I can render "" (empty string) and not break anything?

chariot-client/src/ui/fonts.rs
```diff
@@ -0,0 +1,15 @@
...
+pub const PRIMARY_FONT: FontSelection = FontSelection {
+    source: FontSource::SystemFont {
+        font_name: "ArialMT",
+    },
+    point_size: 32,
+};
+
+pub const PLACEMENT_FONT: FontSelection = FontSelection {
+    source: FontSource::FileFont {
+        file_path: "PressStart2P-Regular.ttf",
```
trulyronak, line 12: Is this path in core?

trulyronak (commented):
Super excited for this - am curious if you can organize string drawable to be within the ui mod though

If thats not possible / you have a reason against that, I'll approve so you can merge

chariot-client/src/ui/fonts.rs
KNOXDEV: Its whatever is defined as the resources path in the configuration, which currently defaults to chariot-core/resources

PR #70, opened 2022-05-24
added multiple player support + colored boxes

trulyronak:
- fixes bug where you couldn't connect more than two clients (it was literally just the player amount was set to 2, this is going to be the last change I make to settings I swear I'm sorry @KNOXDEV )
- renders a different colored rectangle for each player
- fixes bug where player 4's chair rendered in the wrong spot


chariot-core/src/settings.rs
```diff
@@ -46,7 +46,7 @@ impl Settings {
             .set_default("server_address", "127.0.0.1")?
             .set_default("ws_server_port", "0.0.0.0:2334")?
             .set_default("server_tick_ms", 30)?
-            .set_default("player_amount", 2)?
+            .set_default("player_amount", 4)?
```
TheCrossBoy, line 49: LOL

TheCrossBoy (approved):
thank you ronak, very cool

chariot-core/src/settings.rs
trulyronak: BIG

PR #71, opened 2022-05-25
added query set ips, standings, and redirects

trulyronak:
With this update, 
- [X] You can go to `ipaddress?ip=anyip`, making it easy to generate qr codes
- [X] It now auto sets the ip address to be your public ip address (for most local dev, you'll want to use the local ip (127.0.0.1), but if you're on the lab computers it should work


KNOXDEV (approved):
no issues, good work

PR #72, opened 2022-05-25
audio integration

KNOXDEV:
The goal of this PR is to integrate @potor10's brilliant audio framework with the main codebase. We will not exhaustively create every planned audio feature in this PR, but we'll try to get at least one of each major type of audio working so other PR's can expand on it asyncronously. 

- [x] Get music playing on the title screen, chairacter selection, and in game.


KNOXDEV:
Spacial ambient requires a certain level of integration between the map development and the client side code.
In light of this my non-executive decision is to backlog it for the time being, since it's not important for the demo IN 7 DAYS

chariot-client/src/audio/thread/mod.rs
```diff
@@ -1,211 +1,204 @@
...
-  source: AudioBuffer,
-  sink: AudioSinkType,
-  src_opt: SourceOptions,
+    time_start: SystemTime,
+    volume: f32,
+    pitch: f32,
+    source: AudioBuffer,
+    sink: AudioSinkType,
+    // basically, a thread safe duration to set as the fade out
+    fade_out: Arc<Mutex<Option<Duration>>>,
```
alistairgray42, line 30: jeez louise

chariot-client/src/application.rs
```diff
@@ -106,6 +126,12 @@ impl Application {
                     println!("Loading map {}!", map);
                     self.graphics.load_map(map);
                     self.game.signal_loaded();
+                    self.music_manager.play_cf(
```
alistairgray42, line 129: Nice clean easy-to-understand API: sweet!

chariot-client/src/audio/thread/context/mod.rs
```diff
@@ -2,19 +2,19 @@ use rodio::{OutputStream, OutputStreamHandle};
 
 // Audio Context
 pub struct AudioCtx {
-  pub _stream: OutputStream,
-  pub stream_handle: OutputStreamHandle
+    pub _stream: OutputStream,
```
alistairgray42, line 5: Might this be useful in the future? Even if it isn't, it's probably not worthy of a slot in the struct if we're not using it right now

alistairgray42 (approved):
Yeah, this overall feels excellent - nice job!

chariot-client/src/audio/thread/context/mod.rs
KNOXDEV: the reason it needs to exist is because if we drop it or it goes out of scope, the output stream will be destroyed and the music will stop.

PR #73, opened 2022-05-25
release builds, maybe

KNOXDEV:
automated packaged, optimized builds for demo day.

triggered via pushes to main or pr merges


KNOXDEV:
https://github.com/actions/upload-artifact#where-does-the-upload-go for information on how to download artifacts produced from this

KNOXDEV:
Important note:

1. This PR moves all assets related to running the game that weren't ALREADY in `resources` and puts them in the resource folder.
2. This PR moves the resource folder as a whole from `chariot-core/resources` to just `resources` in the root.

KNOXDEV:
### how do I use the artifacts produced from this?

Here is an example: https://github.com/ucsd-cse125-sp22/cse125-sp22-group1/actions/runs/2387742700

You should be able to download the client and server artifacts, unzip them, _download the maps from the chaired drive_, and you should be able to run both the client and server independently. No rust installation or assets management required.

They come with barebones `config.yaml` that you can use to configure the ip addresses quickly, along with pretty much everything else defined in `GLOBAL_CONFIG`.

Right now only building for Windows.

.github/workflows/release.yml
```diff
@@ -0,0 +1,54 @@
...
+  push:
+    branches:
+      - main
+
+name: Build Release - Win64
+
+jobs:
+  client_build:
+    name: Chariot Client Release Builds
+    runs-on: windows-latest
```
trulyronak, line 12: Can you also make it produce builds for Mac and Linux :owo:

trulyronak (approved):
Aaron I love you this is hype 

Can you make it build for Mac and Linux too?

.github/workflows/release.yml
KNOXDEV: yes but that will be another PR for another day

trulyronak: makes sense!!

PR #74, opened 2022-05-26
embedded resource management

KNOXDEV:
let use this to include shaders https://doc.rust-lang.org/std/macro.include_str.html
and this to include audio / ui assets https://doc.rust-lang.org/std/macro.include_bytes.html
into the binary so if they are missing it will throw an error at compile time instead of runtime 
we probably wanna still import the track at runtime since its FAT


alistairgray42 (approved):
All makes sense to me!

PR #75, opened 2022-05-26
Implement a bunch of interactions

alistairgray42:
This implements a bunch of server-side-only interactions (which are the easy ones to implement). Sources for the interactions are #ideas in Discord and our Notion page; lmk if i've missed any.

There are four categories of interactions treated in here:
- Messing with player controls only
- Messing with physics stats only
- Persistent physics things that don't fall into either of those categories (only a couple of these for now)
- One-time events

Each category of `AudienceAction` is handled separately, should be pretty clear where + how.

Also adds a bunch of questions for these actions; please do change the prose of prompts/options on those if you'd like to.


trulyronak (commented):
If theres a way to give staticky defined values to enumerate (I think we can do this via impl methods), I think we should nuke questions.Yaml and just store that data in the enum

alistairgray42:
> If theres a way to give staticky defined values to enumerate (I think we can do this via impl methods), I think we should nuke questions.Yaml and just store that data in the enum

#74 is going to nuke `questions.yaml` and put that data in code anyways; there also won't exactly be bijections between questions, audience actions, and physics changes (e.g. maybe we'll reuse the same actions in different questions, and we'll definitely have some audience actions that don't do physics changes), so there'll have to be that three-way distinction anyways.

chariot-core/src/questions.rs
```diff
@@ -14,22 +14,160 @@ pub struct QuestionOption {
...
-            },
-            QuestionOption {
-                label: "Right".to_string(),
-                action: AudienceAction::NoRight,
-            },
-        ],
-    }];
+    pub static ref QUESTIONS: Vec<QuestionData> = vec![
+        QuestionData {
+            prompt: "Q1 Turning is overrated. Which direction should we ban?".to_string(),
```
trulyronak, line 52: Why label? I wanted to have the questions randomize

chariot-core/src/settings.rs
```diff
@@ -84,6 +85,8 @@ impl Settings {
             .set_default("max_car_spin", 0.1)?
             // How hard we should bounce off the walls (1.0 = as fast as we were initially going)
             .set_default("wall_bounciness", 3.0)?
+            // How hard we shoudl bounce off other players (1.0 = real-world physically accurate)
```
trulyronak, line 88:
```suggestion
            // How hard we should bounce off other players (1.0 = real-world physically accurate)
```

chariot-server/src/game/interactions.rs
```diff
@@ -0,0 +1,136 @@
...
+            let shuffle_order: [usize; 4] = [
+                [1, 0, 3, 2],
+                [1, 2, 3, 0],
+                [1, 3, 0, 2],
+                [2, 0, 3, 1],
+                [2, 3, 0, 1],
+                [2, 3, 1, 0],
+                [3, 0, 1, 2],
+                [3, 2, 1, 0],
+                [3, 2, 0, 1],
```
trulyronak, line 112: i added a shuffle method when I added audience participation (https://docs.rs/shuffle/latest/shuffle/)
Screen Shot 2022-05-27 at 10 28 47 PM

chariot-core/src/questions.rs
alistairgray42: oh so true, good point

chariot-server/src/game/interactions.rs
trulyronak: @alistairgray42 can you make it so the questions auto shuffle each game?

trulyronak (approved):
lgtm

PR #78, opened 2022-05-29
Fix bug where you couldn't get ramped while going fast

alistairgray42:
As per title and commit (bug was that you'd collide off of ramps if you were going too fast when running into them, even from the correct side)


trulyronak (approved):
Abs of steel

PR #83, opened 2022-05-31
Interactions bugs

alistairgray42:
this is gonna collect and fix a whole bunch of small physics/interactions bugs (so far, at least, one bug per commit, so go commit-by-commit for the full experience)

### DIFF
--- a/chariot-server/src/game/mod.rs
+++ b/chariot-server/src/game/mod.rs
@@ -5,6 +5,7 @@ use std::time::{Duration, Instant};
 
 use chariot_core::player::choices::{Chair, PlayerChoices, Track};
 use chariot_core::player::lap_info::Placement;
+use chariot_core::player::player_inputs::PlayerInputs;
 use chariot_core::player::{lap_info::LapInformation, player_inputs::InputEvent, PlayerID};
 use glam::DVec3;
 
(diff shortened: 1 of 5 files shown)


chariot-server/src/game/mod.rs
```diff
@@ -24,7 +25,7 @@ use crate::physics::ramp::RampCollisionResult;
 use crate::progress::get_player_placement_array;
 
 use self::interactions::{
-    get_physics_change_from_audience_action, get_stats_change_from_audience_action,
+    get_physics_change_from_audience_action, get_stats_changes_from_audience_action,
```
TheCrossBoy, line 28: nice

chariot-server/src/chairs.rs
```diff
@@ -17,7 +17,7 @@ use crate::physics::player_entity::PlayerEntity;
 
 // These numbers are completely random guesses btw
 fn get_starting_position_from_player_number(player_number: PlayerID) -> DVec3 {
-    return DVec3::new(0.0, 1.0, 20.0 * (1.5 - player_number as f64));
+    return DVec3::new(21.5 + 2.0 * (1.5 - player_number as f64), 1.0, 65.0);
```
TheCrossBoy, line 20: Ideally we should make this a few objects in Blender, then we can place people based on what place they came in

TheCrossBoy (approved):
this is in fact code!

PR #89, opened 2022-06-01
Add sound effects for menu interactions

alistairgray42:
as per title

i make no pretensions as to be a good judge of sound effects, so mreh if you'd like others feel free to suggest :)


trulyronak (approved):
this is all stuff i'd rather just merge asap

PR #91, opened 2022-06-02
Update chairacter select, create final standings

alistairgray42:
As per title, this modifies two menus:
- Adds chairacter descriptions to the chairacter select screen
- Creates a final placement screen for after the main game


trulyronak (approved):
lgtm 🚀

PR #93, opened 2022-06-02
fullscreen and window title and icon

KNOXDEV:
- [x] set fullscreen on start, config.yaml entry for it
- [x] window title
- [x] window icon
- [x] hotkey to toggle fullscreen during runtime


trulyronak (approved):
based 🚀

KNOXDEV:
Okay two issues:
1. starting the game as full screen or not works fine, but toggling after launch it will not resize the rendering canvas.
2. text rendering does not scale because it uses dpi

chariot-client/src/assets/ui/mod.rs
```diff
@@ -4,6 +4,9 @@ use chariot_core::player::choices::Chair;
 pub const HOME_BACKGROUND: &[u8] = include_bytes!("homebackground.png");
 pub const CHAIR_SELECT_BACKGROUND: &[u8] = include_bytes!("chair-select/background.png");
 
+// icon
+pub const ICON: &[u8] = include_bytes!("icon.png");
```
trulyronak, line 8: take: for no real purpose, I want a type alias called "IMAGES"

PR #100, opened 2022-06-03
countdown

KNOXDEV:
Basic structure is done but needs some polish still. Might add SFX as well.


KNOXDEV:
https://user-images.githubusercontent.com/5607614/171791664-083cd271-1639-43c1-a887-bbfa60f5a950.mp4

no SFX rn but I think that can be a separate PR

if we have the time we might also wanna juice this up a bit

trulyronak (approved):
i love it

PR #103, opened 2022-06-03
Fix progress tracking

alistairgray42:
or at least that's the plan!


chariot-server/src/game/mod.rs
```diff
@@ -647,46 +648,62 @@ impl GameServer {
                     .iter()
                     .enumerate()
                     .all(|(player_num, player)| {
-                        player.lap_info.finished || player_num >= self.connections.len()
+                        if let PlayerProgress::Finished { .. } = player.placement_data {
```
alistairgray42, line 657: a `matches!` and an AND would probably be more elegant here

TheCrossBoy (approved):
imagine approving your own code wtf

PR #120, opened 2022-06-03
adds qr code to game

trulyronak:
so audience members can connect more easily!!


trulyronak:
Screen Shot 2022-06-03 at 1 53 53 PM

this is what it looks like

this QR code goes to [https://cse125.ucsd.edu/2022/cse125g1/redirect](https://cse125.ucsd.edu/2022/cse125g1/redirect), which is instantly editable.

yes this might seem like overkill but it'll make it really seamless for the audience to join

KNOXDEV:
i resolved these conflicts in gh

PR #122, opened 2022-06-03
SHADOW MAP FILTERING MERGE MERGE MERGE

nickwn:
pls test on lab computers first though


KNOXDEV:
lemme test, we'll have both build artifacts on standby if it ends up not working

